PR #7, opened 2026-06-10
Fix wrong instruction mappings in the VMX decode tables

canersaka:
The mnemonics were mapped to the wrong codes, I cross referenced them with the Cell Architecture/Power ISA Manuals

 - vmaxfp (1034) decoded as vminfp; vminfp was 1034 instead of 1098
  - vcfsx/vcfux at 394/330 instead of 842/778 (swapped with vrefp/vrsqrtefp)
  - vmaxub/uh/uw were 256/320/384 instead of 2/66/130
  - the entire vpk* pack family permuted
  - vsumsws was 1032 instead of 1928; vsum4sbs/vsum4shs swapped
  - vcmpgts{b,h,w}/vcmpgtu{b,h,w} signed/unsigned compares swapped
  - Added op31_x*: lvlx (519), lvrx (551), lvlxl (775), ldbrx (532).
  - 
These decode silently. The lifter generates valid C for the wrong operation, so a recompiled game just computes wrong values (my game had around 15k affected instructions; vcfsx alone appears 9,728 times). Tested by decoding real instruction words from my game's executable plus hand-built encodings: 18/18 correct after the fix.

PR #8, opened 2026-06-10
Add vmx lifter handlers

canersaka:
Note: this branch includes the decode-fix commit from #7, since the handlers depend on those instructions decoding correctly. If PR #7 merges first, this PR reduces to just the handler additions.

These instructions had no translation handler and fell through to the lifter's TODO catch-all. In my game, addc/subfc alone were 86% of all unlifted instructions (63k of 73k), the recompiled output couldn't run without them.

Added (from the PowerPC/AltiVec manuals):

addc/subfc with carry-out into XER[CA] (addc: carry when the unsigned add wraps; subfc: carry when rB >= rA unsigned)
The unaligned vector loads lvlx/lvrx/lvlxl (load the aligned 16-byte block, then shift by the in-block offset, deliberately not implemented as lvx aliases) and ldbrx
vsrw, vsububs, vsum2sws, vupkhsh, vrfim
vand had an implementation but was missing from the dispatch list, so it never ran
Fixed swapped operands in vrefp/vrsqrtefp

Progress logging during the lift (quality of life)

Testing: compiled the new handlers with MSVC and asserted outputs against hand-computed values (9/9 pass). Together with the decode fixes, my game's unlifted-instruction count went from 73,219 to nearly zero.

### DIFF
--- a/tools/ppu_disasm.py
+++ b/tools/ppu_disasm.py
@@ -771,6 +771,22 @@ def decode(insn: int, addr: int = 0) -> Instruction:
             result.mnemonic = "lvsr"
             result.operands = f"v{rd}, r{ra}, r{rb}"
             return result
+        if xo_full == 519:  # lvlx (Cell ext: Load Vector Left Indexed)
+            result.mnemonic = "lvlx"
+            result.operands = f"v{rd}, r{ra}, r{rb}"
+            return result
+        if xo_full == 551:  # lvrx (Cell ext: Load Vector Right Indexed)
+            result.mnemonic = "lvrx"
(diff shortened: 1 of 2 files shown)


sp00nznet:
🙏 Thank you, @canersaka! This landed on `master` in 1ee110d (your commit — authorship preserved) as part of the **v0.6.1 "Many Hands"** release: it added the missing VMX lifter handlers (addc/subfc carry, lvlx/lvrx/lvlxl, ...). You're credited in [CONTRIBUTORS.md](https://github.com/sp00nznet/ps3recomp/blob/master/CONTRIBUTORS.md) and the README changelog.

PR #9, opened 2026-06-11
fix NID computation: correct suffix constant and use little-endian di…

canersaka:
Fix NID computation in nid_database.py

While resolving imports for Yakuza port, nid_database.py matched 0 of 354 import NIDs. Two bugs: the hash suffix constant was truncated or corrupted (12 bytes, wrong from byte 8 — include/ps3emu/nid.h has the correct 16-byte value), and the first 4 digest bytes were unpacked big-endian instead of little-endian.

Verification: sha1("cellSysmoduleLoadModule" + suffix)[0:4] little-endian gives 0x32267A31, which matches the NID in a real EBOOT import table. With the fix, 343/354 of my game's import NIDs resolve to names.

### DIFF
--- a/tools/nid_database.py
+++ b/tools/nid_database.py
@@ -32,10 +32,12 @@ import sys
 # ---------------------------------------------------------------------------
 
 # The suffix that Sony appends before hashing.
-# This is the well-known suffix used for PS3 NID generation.
+# This is the well-known 16-byte suffix used for PS3 NID generation
+# (verified against a real EBOOT import table and RPCS3's ppu_generate_id;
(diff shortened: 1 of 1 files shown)


sp00nznet:
🙏 Thank you, @canersaka! This landed on `master` in a9a05ce (your commit — authorship preserved) as part of the **v0.6.1 "Many Hands"** release: it fixed NID computation (16-byte suffix + little-endian) — foundational. You're credited in [CONTRIBUTORS.md](https://github.com/sp00nznet/ps3recomp/blob/master/CONTRIBUTORS.md) and the README changelog.

PR #11, opened 2026-06-11
Lifter fixes + Huge Speed Improvement through Parallelization

canersaka:
Improvements I made while trying to push Yakuza recomp further, found from pushing a 22 MB / ~45k-function game through the lifter. I can separate the fixes if needed, overall it just sped up the lifter process immensely, and fixed some bugs.

I was trying to relift multiple times, and each one took multiple hours while the program ran on a single core, so I parallelized it. Now it does the same work in less than a few minutes.

Parallel lifting: --jobs N (default: CPU count) separates the per-function translation out to a process pool; workers re-disassemble only their own byte ranges so nothing large crosses boundaries. Discovery and emission stay serial, and --jobs 1 is the original path, and not touched. A full lift went from hours to minutes for me. Also added progress/ETA output to the lift and emit phases.

Record-form CR0 bug: Dot-form integer instructions (add., andi., adde., ...) are supposed to set CR0 from the result. The lifter did the arithmetic but left CR0 stale, so the next conditional branch went whichever way the previous compare said. Now emitted generically for integer dot-forms; stwcx./stdcx. (set CR0 themselves), FP (CR1) and VMX (CR6) forms excluded. Also adde./addze./subfe. didn't match their handlers at all. Only the dotless spellings were checked.

Context layout bug: The generated ppu_context had drifted from runtime/ppu/ppu_context.h (xer width, fpscr/vscr order, tail fields). The same pointer flows from recompiled code into runtime syscall handlers, so any handler touching a drifted field silently reads or writes the wrong bytes. The generated struct now matches the runtime field-for-field.

Split output files: My game's generated source was 19 million lines, and MSVC fails any source over around 16 million lines. Output is now chnked into ppu_recomp_NNN.cpp (~600k lines each, cut at function boundaries). By doing this MSVC does not fail and the chunks can compile in parallel and rebuild incrementally. The function table also gets external linkage plus a declaration in the generated header (it was static, so game projects' call dispatchers couldn't use it).

Testing: parallel output verified byte-for-byte identical to serial (134k generated lines compared); CR0 emission spot-checked on hand-built encodings of andi./add./subfc. (and confirmed plain add gets no CR0 write); chunked output verified to place the function table only in the final chunk; the full pipeline (lift → compile with MSVC → link → run) works end-to-end on my game, which now boots through to its C++ constructor pass.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -16,6 +16,7 @@ import json
 import os
 import re
 import sys
+import time
 from dataclasses import dataclass, field
 
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


sp00nznet:
🙏 Thank you, @canersaka! This landed on `master` in dd9bda8 (your commit — authorship preserved) as part of the **v0.6.1 "Many Hands"** release: it parallelized the lifter (--jobs) + chunked output + record-form CR0. You're credited in [CONTRIBUTORS.md](https://github.com/sp00nznet/ps3recomp/blob/master/CONTRIBUTORS.md) and the README changelog.

PR #12, opened 2026-06-11
Fix sradi decode for shift amounts >= 32

canersaka:
Why: sradi encodes its 6-bit shift with the top bit in instruction bit 30, inside the field read as a 10-bit XO. Shifts under 32 read as XO 826 (which was in the table); shifts of 32+ read as 827 (missing), so those instructions didn't decode at all.

Change: added 827 and composed the shift as (bit30 << 5) | sh.

Testing: 0x7C64FE76 decodes as sradi r4, r3, 63; sub-32 encodings unchanged (sradi r4, r3, 5 still correct).

### DIFF
--- a/tools/ppu_disasm.py
+++ b/tools/ppu_disasm.py
@@ -536,14 +536,18 @@ def decode(insn: int, addr: int = 0) -> Instruction:
         x_shift = {
             24: "slw", 536: "srw", 792: "sraw", 824: "srawi",
             27: "sld", 539: "srd", 794: "srad", 826: "sradi",
+            827: "sradi",
         }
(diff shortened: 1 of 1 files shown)


sp00nznet:
🙏 Thank you, @canersaka! This landed on `master` in ffc6fc5 (your commit — authorship preserved) as part of the **v0.6.1 "Many Hands"** release: it fixed sradi decode for shifts >= 32. You're credited in [CONTRIBUTORS.md](https://github.com/sp00nznet/ps3recomp/blob/master/CONTRIBUTORS.md) and the README changelog.

PR #13, opened 2026-06-11
Improve function detection: seed from the .opd table

canersaka:
Why: my recompiled game crashed on its first call through a function pointer, the target function was never detected, so it was never lifted. Found that PS3 executables carry a .opd section listing a descriptor for every address-taken function (every C++ virtual, every callback). My game has 21,893 of them; exactly 1 matched a detected function start. The prologue scan anchors on mflr r0 and registers starts late when the frame is set up first, and pointer-only functions are invisible to bl following.

Changes:
Seed starts from .opd, located by shape (data-segment pairs of valid code address + repeating TOC value), since section names are often stripped. The ELF entry point descriptor is seeded too.
Seeded functions end at their first blr, bounded by the next known start.
Only trust bl/branch targets found inside already-detected functions , executable segments contain read-only data, and data decoded as instructions yields phantom branch targets; real code never branches into data.

Clip overlapping ranges so each byte belongs to one function.

Binary search instead of a linear scan in the branch-target pass (Multiple minutes is now less than one second on my game; whole tool now is now just a few seconds).

Testing: the tool now self-checks and reports that every .opd address is a function start. On my game: 30,721 became 45,425 functions, and the recompiled build went from crashing on the first virtual call to running the entire C++ global constructor pass.

### DIFF
--- a/tools/find_functions.py
+++ b/tools/find_functions.py
@@ -10,14 +10,14 @@ Analyses a PPU ELF or raw binary for function entry/exit points by:
   - Detecting standard PPU prologues (mflr r0; stw/std r0, X(r1); stwu/stdu r1, -Y(r1))
   - Detecting epilogues (lwz/ld r0, X(r1); mtlr r0; blr)
   - Following bl (branch-and-link) targets to discover called functions
-  - Closing inter-function gaps so no text bytes go unattributed (stray data
-    lifts harmlessly as .word)
+  - Clipping overlapping ranges so every byte belongs to at most one function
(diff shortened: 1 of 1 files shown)


sp00nznet:
🙏 Thank you, @canersaka! This landed on `master` in 5a45971 (your commit — authorship preserved) as part of the **v0.6.1 "Many Hands"** release: it seeded function detection from the .opd table. You're credited in [CONTRIBUTORS.md](https://github.com/sp00nznet/ps3recomp/blob/master/CONTRIBUTORS.md) and the README changelog.

PR #14, opened 2026-06-12
runtime: implement the sys_vm syscall family (300-313)

canersaka:
The sys_vm virtual-memory syscalls were unimplemented; sys_vm_memory_map (300) returns the mapped base via a guest u32 out-pointer, so callers got ENOSYS plus an unwritten out-param and crashed dereferencing NULL. Yakuza calls it during ctor-pass init (asks for a 128 MB arena) and died on the first arena-header write.

Call contract verified against the calling code in the game binary (rounds vsize to 32 MB / psize to 64 KB itself, reads back a u32, touches it via 306, uses it as a heap base); RPCS3 sys_vm.cpp used as read-only cross-check. Mappings are bump-allocated from a dedicated guest window at 0x60000000 and fully committed, which trivially satisfies the paging-hint syscalls (lock/touch/flush/... are no-ops returning CELL_OK).

sys_vm_test (311) and sys_vm_get_statistics (312) are intentionally left on the loud default ENOSYS handler: both write out-structs whose layout is not yet verified from a primary source.

Boot-validated: map succeeds and the boot advances past the old crash.

### DIFF
--- a/runtime/syscalls/lv2_register.c
+++ b/runtime/syscalls/lv2_register.c
@@ -19,6 +19,7 @@
 #include "sys_event.h"
 #include "sys_timer.h"
 #include "sys_memory.h"
+#include "sys_vm.h"
 #include "sys_fs.h"
 #include "ps3emu/spu_fallback.h"
 #include "sys_event.h"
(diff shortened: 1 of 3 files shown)


sp00nznet:
🙏 Thank you, @canersaka! This landed on `master` in e40896f (your commit — authorship preserved) as part of the **v0.6.1 "Many Hands"** release: it implemented the sys_vm syscall family (300-313). You're credited in [CONTRIBUTORS.md](https://github.com/sp00nznet/ps3recomp/blob/master/CONTRIBUTORS.md) and the README changelog.

PR #16, opened 2026-06-12
 libs: lv2-correct sys_lwmutex destroy semantics + slot locking

canersaka:
sys_lwmutex_destroy returned OK unconditionally; real lv2 returns ESRCH
for a dead mutex and EBUSY for a held one, and games' libc teardown paths
branch on those codes. Also locks slot allocation (creates from multiple
guest threads raced). First commit adds the create/lock diagnostics used
to find this.
Evidence: Yakuza's libc heap finalization calls destroy on the same mutex
twice on real hardware and depends on the second returning ESRCH; with
unconditional OK its allocator state corrupts.

### DIFF
--- a/libs/system/sysPrxForUser.c
+++ b/libs/system/sysPrxForUser.c
@@ -244,10 +244,13 @@ s32 _sys_tolower(s32 c) { return tolower(c); }
  * Lightweight mutex
  * -----------------------------------------------------------------------*/
 
+extern u8* vm_base;  /* for guest-address diagnostics in the boot log */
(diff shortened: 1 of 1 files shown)


sp00nznet:
🙏 Thank you, @canersaka! This landed on `master` in d4c79d8 (your commit — authorship preserved) as part of the **v0.6.1 "Many Hands"** release: it made sys_lwmutex_destroy lv2-correct (ESRCH/EBUSY) + slot locking. You're credited in [CONTRIBUTORS.md](https://github.com/sp00nznet/ps3recomp/blob/master/CONTRIBUTORS.md) and the README changelog.

PR #18, opened 2026-06-12
libs: big-endian guest structs in cellVideoOut; fix CellGcmContextData layout

canersaka:
Out-param structs live in guest memory, so multi-byte fields must be written big-endian. GetResolution wrote host-endian width/height - on x86 the guest reads 720x480 as 53250x57345 and sizes its display buffers from garbage. Fixes GetResolution, GetState.displayMode, GetConfiguration.pitch, and reads config->pitch as BE; libs-internal callers keep a host-endian helper. Also corrects CellGcmContextData: callback was at offset 0, but compiled SDK inline code (verified in a commercial EBOOT) and the gcm ABI place begin/end/current/callback with callback at +0xC.

Evidence: before, Yakuza's display-buffer setup produced 53250x57345; after, 720p configures coherently (pitch 5120) and flips run with correct buffer ids.

### DIFF
--- a/libs/video/cellGcmSys.h
+++ b/libs/video/cellGcmSys.h
@@ -93,12 +93,15 @@ typedef struct CellGcmControl {
 /* GCM context data — the game reads/writes this to submit RSX commands.
  * On PS3, gCellGcmCurrentContext points to one of these.
  * All pointer fields are guest addresses (u32) in the recomp.
- * Layout verified from compiled PS3 SDK inline functions: callback is at offset 0. */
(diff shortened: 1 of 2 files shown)


sp00nznet:
🙏 Thank you, @canersaka! This landed on `master` in 54fce99 (your commit — authorship preserved) as part of the **v0.6.1 "Many Hands"** release: it fixed big-endian guest structs in cellVideoOut + CellGcmContextData layout. You're credited in [CONTRIBUTORS.md](https://github.com/sp00nznet/ps3recomp/blob/master/CONTRIBUTORS.md) and the README changelog.

PR #21, opened 2026-06-21
ppu_lifter: XER[CA] for shift-algebraic ops + mtcrf field mask + rotate sh==0 guard

canersaka:
sraw/srawi/srad/sradi now set XER[CA] (rS<0 AND any 1-bits shifted out) via ppc_sraw/ppc_srad so adde/addze/subfe consumers are correct, and clamp shifts >= width (UB in C). mtcrf updates only the CRM-selected CR fields instead of overwriting all of CR. rlwinm/rlwimi/rotl64 guard the sh==0 case (shift-by-width is UB). Verified in the Yakuza port boot.


sp00nznet:

PR #23, opened 2026-06-21
libs: real cellAudio port implementation

canersaka:
PortOpen read the BE guest param host-native (nChannel came back 0x08...00 not 8 -> CELL_AUDIO_ERROR_PARAM -> mixer teardown). Byteswap the param; allocate the ring buffer + read index in guest VM; write CellAudioPortConfig BE at real offsets; advance the read index per block; push the period event via a new sys_event_find_queue_by_key helper.

The audio buffer region is placed at guest 0x50000000; if that collides with another title's map, vm_commit fails and audio degrades gracefully rather than crashing. **May need checking for other titles.**
 Verified in the Yakuza port.

### DIFF
--- a/libs/audio/cellAudio.c
+++ b/libs/audio/cellAudio.c
@@ -17,6 +17,14 @@
 #include <string.h>
 #include <stdlib.h>
 #include <math.h>
+#include "../../runtime/memory/vm.h"   /* vm_commit, vm_to_host, vm_base */
+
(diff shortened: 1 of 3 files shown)


canersaka:
Superseded by #54: upstream independently landed the buffer/endianness half of this, so #54 re-submits just the still-missing event delivery against the current tree. Closing this one.

PR #25, opened 2026-06-26
sys_rwlock: return EDEADLK on writer self-relock and EPERM on bad unlock

canersaka:
The Win32 write-lock path used a non-recursive SRWLock with no ownership tracking, so two lv2-defined error conditions were mishandled:

- A thread that already holds the write lock calling sys_rwlock_wlock again would AcquireSRWLockExclusive on a lock it owns -> hard host deadlock. lv2 returns CELL_EDEADLK.
- sys_rwlock_wunlock by a thread that is not the writer (or with no writer) called ReleaseSRWLockExclusive, which is undefined behaviour for an SRWLock not held exclusively by the caller. lv2 returns CELL_EPERM.

Track the owning thread (writer_tid, set under the lock, using the same ctx->thread_id convention as sys_mutex's owner_tid) and check it in wlock/trywlock (EDEADLK) and wunlock (EPERM). The pthread path is unchanged (glibc rwlocks already return EDEADLK/EPERM).

### DIFF
--- a/runtime/syscalls/sys_rwlock.c
+++ b/runtime/syscalls/sys_rwlock.c
@@ -171,8 +171,13 @@ int64_t sys_rwlock_wlock(ppu_context* ctx)
         return (int64_t)(int32_t)CELL_ESRCH;
 
 #ifdef _WIN32
+    /* A thread that already holds the write lock re-acquiring it would deadlock
+     * the (non-recursive) SRWLock. Real lv2 returns CELL_EDEADLK. */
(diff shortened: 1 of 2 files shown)


sp00nznet:
Incorporated into `master` in [`f30d03d`](https://github.com/sp00nznet/ps3recomp/commit/f30d03d) (cherry-picked, v0.6.5) and credited in the README changelog + CONTRIBUTORS.md. Thanks! Closing as landed.

PR #26, opened 2026-06-26
ppu_lifter: correct XER[CA] for subfe/subfme/addme

canersaka:
The carry-out (XER[CA]) for the subtract/add-extended ops was wrong:

- subfe set CA via `result <= RB && ca` -- wrong for ~42/200 random operand pairs vs a 65-bit reference.
- addme set CA via `result >= RA` -- wrong for ~10/20.
- subfme never wrote XER[CA] at all, leaving it stale.

These ops are RA + RB(or ~0) + CA, and CA is the carry-out of that 65-bit add. Compute it exactly with the two-step ADC model RPCS3 uses in add64_flags (s1 = a+b, co = s1<a; result = s1+ca, co |= result<s1), matching the instruction bodies: subfe = ~RA + RB + CA, addme = RA + ~0 + CA, subfme = ~RA + ~0 + CA. The result value is unchanged; only CA is corrected. subfze was already correct and is left untouched.

A wrong CA silently corrupts any multi-word carry chain (a following adde/subfe/addze reads the bad carry) -- wrong arithmetic, not a crash.

Verified: re-simulated CA and result for all four ops against a 65-bit reference (0 mismatches); ppu_lifter.py parses clean.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -1200,10 +1200,12 @@ class PPULifter:
 
         if mn in ("subfe", "subfe."):
             rd, ra, rb = _reg_idx(ops[0]), _reg_idx(ops[1]), _reg_idx(ops[2])
+            # XER[CA] = ADC carry-out of (~RA + RB + ca), per PowerISA / RPCS3 add64_flags.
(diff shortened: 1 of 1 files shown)


sp00nznet:

PR #27, opened 2026-06-26
cellPad: write CellPadData / CellPadInfo2 out-params big-endian

canersaka:
The pad out-params were written host-endian, so the guest (which reads them big-endian) saw byte-swapped values: CellPadData.len plus every button/ analog/sensor halfword in button[], and all the u32 fields of CellPadInfo2 (max_connect, now_connect, and the port_status / port_setting / device_capability / device_type arrays).

Byte-swap them on write via ps3_bswap16/32. CellPadData.button is u16[]; CellPadInfo2's fields are u32 (all be_t<u32> in RPCS3's cellPad.h). Single- byte fields are left as-is.

### DIFF
--- a/libs/input/cellPad.c
+++ b/libs/input/cellPad.c
@@ -11,6 +11,7 @@
  */
 
 #include "cellPad.h"
+#include "ps3emu/endian.h"   /* ps3_bswap16/32: CellPadData/Info2 are guest big-endian */
 #include <stdio.h>
 #include <string.h>
 #include <math.h>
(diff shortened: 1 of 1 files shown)


sp00nznet:
Closing as **obsolete** — the bug is fixed, but by a different route, so this patch no longer applies cleanly to intent.

You diagnosed it correctly: the pad out-params were reaching the guest byte-swapped. What's changed is the mechanism. `cellPad.c` has since moved off host pointers entirely onto the guest-EA accessors:

```c
vm_write32((unsigned long long)ea + 0, (unsigned int)data->len);        /* len (s32) */
vm_write16((unsigned long long)ea + 4 + i * 2, data->button[i]);        /* button[] */
vm_write32(gaddr + 0x00, s_max_connect);                               /* CellPadInfo2 */
```

…and `vm_write32` byte-swaps on the way out:

```c
void vm_write32(uint64_t a, uint32_t v) { ...
    v = __builtin_bswap32(v); memcpy(vm_base + (uint32_t)a, &v, 4); }
```

So every field this PR targets — `CellPadData.len`, `button[]`, and `CellPadInfo2`'s `max_connect` / `now_connect` / `port_status` / `port_setting` / `device_capability` / `device_type` — is now written big-endian already, and through a path that also does the EA translation the host-pointer version was missing.

Applying this on top would **double-swap** (`ps3_bswap32` then `vm_write32`'s swap) and write through a host pointer instead of the guest EA.

Not a rejection — the endianness call was right, and the same reasoning is what drove the `vm_write` conversion. Just overtaken by it.

Your other 13 PRs landed in #76 and are closed individually with their commits; #65 and #69 are going in now — #65 turned out to be **more** right than its description claimed (details there).

PR #29, opened 2026-06-27
video/gcm: fix cellGcmGetTiledPitchSize ABI and flip/vblank handler dispatch

sp00nznet:
## Summary

Two GCM/RSX bugs found while bringing up *You Don't Know Jack* (BLUS30569) under the recompiler. Both prevented the title's RSX command stream from being issued correctly. With these fixes the title's own recompiled code reaches the FIFO and issues real `NV4097_CLEAR_SURFACE` commands (visible clears on screen via the D3D12 backend).

### 1. `cellGcmGetTiledPitchSize` had the wrong ABI

The real export is `uint32_t cellGcmGetTiledPitchSize(uint32_t size)` — **one** argument, with the aligned tiled pitch returned in `r3`. The previous 2-arg form treated `r4` as an out-pointer and wrote the pitch through it.

At the title's call site `r4` still held the caller's `CellVideoOutResolution*` (left over from the preceding `cellVideoOutGetResolution`), so the write clobbered `width`/`height` with the pitch value. `width` became `0`, and every downstream tile / display-buffer setup computed from garbage.

### 2. Flip/vblank handler callbacks dispatched through a clobbered OPD

Handlers were dispatched by re-resolving the handler OPD at tick time. The title's handler OPDs (`0x530D70/78/80`) have their code word clobbered in guest memory by a later lifted store, so re-resolving yields the TOC base instead of the real entry point.

Fix: capture the resolved `{code, toc}` at registration time and dispatch through a new `ppu_guest_call_ct()` (call-by-resolved-code/toc) in the PPU loader, which runs the real lifted handler regardless of the later OPD corruption.

## Testing

- `ps3recomp_runtime` builds clean (Release).
- `ydkj_boot.exe` links clean end-to-end against the new `cellGcmGetTiledPitchSize` signature.
- At runtime the title issues real `CLEAR_SURFACE` from its own recompiled code.

### DIFF
--- a/libs/video/cellGcmSys.c
+++ b/libs/video/cellGcmSys.c
@@ -109,6 +109,20 @@ static u32 s_flip_handler_opd     = 0;
 static u32 s_vblank_handler_opd   = 0;
 static u32 s_user_handler_opd     = 0;
 static u32 s_second_v_handler_opd = 0;
+
+/* Resolved {code,toc} captured when each handler is registered. The title's
+ * handler OPDs (0x530D70/78/80) get their code word clobbered in guest memory
+ * by a later lifted store, so re-resolving the OPD at tick time yields the TOC
+ * base instead of the real handler. Capture-at-registration + dispatch by
+ * code+toc runs the real lifted handler regardless. */
+extern unsigned vm_read32(unsigned long long);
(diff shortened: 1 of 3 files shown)


sp00nznet:
Already in master. `cellGcmGetTiledPitchSize` now takes the single `size` argument per the SDK prototype (libs/video/cellGcmSys.c:1917) and the flip/vblank handler dispatch went in with it. Closing as superseded — the branch was carrying a stale base.

PR #33, opened 2026-07-02
spu_disasm: fix il negative-immediate double sign-extension

canersaka:
i16 is sign-extended once at field extraction; the il operand formatter extended it again (Python: -4 & 0x8000 is truthy), so every negative il immediate in every SPU lift came out as imm - 0x10000. The same double extension existed in the RI18-branch copy; both corrected.

Validated on the Yakuza: Dead Souls port: all five lifted SPU modules relifted, and the previously-crashing Edge geometry task (il $r17,-4 lifted as -65540 sent its free-list restock to LS 0x3BDC0 instead of 0xBDC0).


sp00nznet:

PR #34, opened 2026-07-02
spu: preferred-slot-only link register and rchcnt count

canersaka:
The SPU writes scalar results into the PREFERRED word (word 0) with the remaining three slots ZERO (RPCS3 v128::from32r): the brsl/brasl/bisl link register and the rchcnt channel count. The lifter splatted both into all four words. Latent wherever only word 0 is consumed (bi $r0 returns, biz gates), which is why it survived boot validation, but it corrupts any code that saves/reloads/operates on the full 128-bit quadword; Sony's SPURS taskset context save does exactly that with the link (lqd/stqd of r0).  Adds spu_link/spu_pref_u32 to spu_helpers.h 
Validated on the Yakuza: Dead Souls port: all five lifted SPU modules relifted + full boot regression pass.


sp00nznet:

PR #35, opened 2026-07-02
ppu_lifter: cntlzw(0) is 32, not undefined __builtin_clz garbage

canersaka:
PowerISA defines cntlzw of zero as 32; the lifter emitted a raw __builtin_clz, which is undefined for zero input (x86 BSR leaves the index unset), and the MSVC compat shim additionally ignored _BitScanReverse's found flag and returned 31 minus an uninitialized index. cntlzd already had the zero guard; cntlzw did not, so every cntlzw of a zero value in a lifted game silently computed garbage. Fixed with a guarded emission (x ? clz : 32) mirroring cntlzd, plus zero-safe MSVC clz/clzll shims.

Found and validated on the Yakuza: Dead Souls port: Sony's cellSpursQueuePushBody gates its "queue was empty, wake the registered task" signal on cntlzw(pending)>>5. The garbage result dropped every SPURS queue wakeup, leaving the CRI audio codec task asleep in WAIT_SIGNAL forever. With the guard the push raises the task signal and the codec dispatches.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -123,12 +123,17 @@ SOURCE_PREAMBLE = """\
  * redefining them is a hard error there -- only polyfill for real MSVC. */
 #ifndef __clang__
 static inline int __builtin_clz(unsigned int x) {
+    /* BSR leaves the index undefined for x==0; PowerPC cntlzw(0) is DEFINED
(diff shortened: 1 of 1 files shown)


sp00nznet:

PR #36, opened 2026-07-02
spu_lifter: bi $r0 reloaded via lqa/lqr is a computed tail jump, not a return

canersaka:
As far as I can tell Sony's SPU code uses `bi $r0` two ways: the standard function return (r0 = link register, restored from the stack via lqd/lqx or never rewritten) and a computed tail jump after reloading r0 from an absolute address -- the SPURS taskset dispatcher launches a task with `lqa $r0, savedContextLr; ... bi $r0`, jumping to the task entry, not back to a caller. Lifting every `bi $r0` as a host `return;` makes such a launch fall through and the dispatcher loop forever. compute_bi_r0_jumps() classifies each site by the nearest preceding r0 writer (lqa/lqr means computed jump, emitted as the existing spu_indirect_branch dispatch; anything else stays a return). The scan is RRR-aware (selb/shufb/mpya/fma-family write bits 21-27, so their raw&0x7F RC source field must not be mistaken for an r0 write). The biz/binz/bihz/bihnz family gets the matching treatment: a conditional branch to an unflagged $r0 is a conditional return (the old unconditional dispatch would land on an unregistered mid-function PC).

Lifting the SPURS taskset policy module from libsre with this flags its launch sites (`2 bi $r0 computed-jump site(s)` in the CLI summary); on the Yakuza: Dead Souls port this classifier is what lets the taskset dispatcher actually launch tasks (the Edge geometry task, the CRI audio codec) instead of spinning. Directly relevant to the YDKJ SPURS task-dispatch work: that launch path ends in exactly this instruction.

### DIFF
--- a/tools/spu_lifter.py
+++ b/tools/spu_lifter.py
@@ -179,6 +179,56 @@ _NO_RT_WRITE = {
 }
 
 
+def _bi_target_reg(insn):
+    """Return the branch-target register of a bi/biz/binz/... insn, or None."""
+    ops = _ops(insn.operands)
+    if insn.mnemonic == "bi" and ops:
+        return _reg(ops[0])
+    if insn.mnemonic in ("biz", "binz", "bihz", "bihnz") and len(ops) >= 2:
+        return _reg(ops[1])
+    return None
+
+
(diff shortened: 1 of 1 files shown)


sp00nznet:

PR #37, opened 2026-07-02
tools: PPU lifter conformance suite + six emission fixes it caught

canersaka:
test_ppu_lift.py tests the lifter's ACTUAL emissions, not a parallel implementation: each case encodes a real instruction word, round-trips it through ppu_disasm.decode (so an encoding mistake reports as ENCODING, never a false lifter failure), asks PPULifter for the C it would emit, and runs all emitted statements in a generated C driver against edge-case register inputs, comparing with expectations computed in Python straight from the PowerISA definitions. A wrong handler compiles and silently computes the wrong thing, and nothing else in the pipeline checks lifted-code semantics before a game misbehaves at runtime; on the Yakuza: Dead Souls port this suite's first run caught six silent miscompile classes in milliseconds.

Its first run against this repo's master found 119 failing checks. This PR fixes the six classes not already covered by an open PR: addi/addis/addic truncated the 64-bit add to 32 bits; addic/subfic never wrote XER[CA] (subfic also computed in 32 bits); mulli/mullw truncated the 64-bit product; slw/srw sign-extended the 32-bit result into the high word where PowerISA zero-extends; neg was a 32-bit negate (neg of INT64_MIN must stay INT64_MIN); eqv was unhandled entirely. The remaining 46 failures are exactly the classes fixed by open PRs #21 (sraw/srawi/sradi XER[CA]) and #26 (subfe/subfme/addme XER[CA]) plus #35 (cntlzw of zero) -- with those three merged locally on top of this branch the suite passes 1163/1163, 0 skipped.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -503,16 +503,25 @@ class PPULifter:
             return f"ctx->gpr[{rd}] = (int64_t)(int32_t)((uint32_t){_imm(ops[1])} << 16);"
 
         if mn == "addi":
+            # Full 64-bit add (PowerISA): truncating to 32 bits corrupts any
+            # 64-bit pointer/counter arithmetic built with addi.
             rd, ra = _reg_idx(ops[0]), _reg_idx(ops[1])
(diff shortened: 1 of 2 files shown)


sp00nznet:

PR #38, opened 2026-07-03
ppu_lifter: mftb/mftbu read a real timebase, not a per-call-site counter

canersaka:
The mftb/mftbu emission gave every call site a private static counter advancing 16667 ticks per read. That is not a clock: it only moves when polled (a loop that reads the timebase twice with work in between measures zero elapsed time, while a tight poll measures fast-forwarded time), every call site disagrees with every other, and none of it relates to sys_time_get_timebase_frequency's correctly-reported 79.8 MHz. Any guest code that computes elapsed time from the timebase -- media pacers, frame throttles, profilers, timeout loops -- computes garbage. mftb/mftbu now read ppu_timebase_now(), a global monotonic counter added to runtime/syscalls/sys_timer.c: host time (QPC on Windows, CLOCK_MONOTONIC elsewhere) scaled to the PS3 timebase with an overflow-safe split multiply, anchored at first use; the generated-source preamble declares the extern, and mftbu returns the high word of the same counter so mftbu/mftb pairs stay coherent.

Found and validated on the Yakuza: Dead Souls port: the CRI media pacer computes its scheduling period from mftb deltas, and against the per-site counter its period stretched about a hundredfold, starving the async-file pump it drives and wedging the title's shader-archive streaming. With the real timebase the pacer runs at its intended rate; full relift plus boot, and the instruction conformance suite (PR #37) stays green.

### DIFF
--- a/runtime/syscalls/sys_timer.c
+++ b/runtime/syscalls/sys_timer.c
@@ -25,6 +25,36 @@ static void ensure_qpc_init(void)
 }
 #endif
 
+/* ---------------------------------------------------------------------------
+ * PPU timebase (mftb/mftbu) -- THE guest clock.
+ *
+ * One global monotonic counter scaled to the PS3 timebase (79.8 MHz),
+ * anchored at first use. The old lifter emission was a per-call-site
+ * static that advanced 16667 ticks PER READ -- not a clock at all: every
+ * guest timing loop (media pacers, throttles, profilers) computed garbage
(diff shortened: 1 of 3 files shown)


sp00nznet:
Incorporated into `master` in [`3480aac`](https://github.com/sp00nznet/ps3recomp/commit/3480aac) (v0.6.4 "Carry the One"). Credited as co-author on the commit and by name in the changelog + CONTRIBUTORS.md. Thanks! Closing as landed.

PR #39, opened 2026-07-03
ppu_lifter: vcmpgtsb/sh/sw/ub/uh/uw handlers (dot forms set CR6)

canersaka:
The vector compare-greater-than family (vcmpgtsb/sh/sw, vcmpgtub/uh/uw, and their dot forms) had no handler and fell through to the TODO catch-all, which skips the instruction entirely: vD silently keeps stale data and a branch on the dot form's CR6 result tests whatever CR6 happened to hold -- silent wrong-data failures, not crashes. Added table-driven emission over the six element types in the same shape as the existing vcmpequ* handlers; dot forms write CR6 per the AltiVec PEM (bit 0, value 8, all lanes true; bit 2, value 2, all lanes false). Like the neighboring VMX handlers the lane math uses the repo's host-order lane convention: byte-width compares are unaffected by lane order and the half/word forms are consistent with the existing vcmpequh/w.

Found and validated on the Yakuza: Dead Souls port: 35+ live sites in its effect/particle math and codec-style compare-pack pipelines were being skipped. Full relift plus boot, and conformance-suite cases (PR #37 harness) covering all-true, all-false, and mixed lanes check both the destination lanes and the CR6 nibble.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -1708,6 +1708,30 @@ class PPULifter:
                     f"uint32_t* d=(uint32_t*)&ctx->vr[{vd}]; "
                     f"for(int i=0;i<4;i++) d[i]=a[i]==b[i]?~0u:0; }}")
 
+        # Vector compare greater-than, signed/unsigned, byte/half/word.
+        # Previously unhandled (fell to the TODO catch-all = silently skipped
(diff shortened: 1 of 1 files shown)


sp00nznet:

PR #43, opened 2026-07-05
RSX 2D/dbgfont rendering via real vertex program

sagemono:
Adds the initial 2D/dbgfont pipeline. Testing and debugging was performed using a homebrew app that utilises a TUI driven by libdbgfont.  

> * NV4097 vertex-program capture; executes the title's real RSX vertex program (D3D12) for pixel-perfect text.
> * Bigger vertex buffer, vsync throttle, per-clear frame reset.
> * cellGcmSys: recycle the FIFO ring on wrap — the command ring never recycled (callback was a no-op) → wedged ~200 frames in. Now a real command-buffer-full callback drains + resets.

Also selects the discrete GPU by default since my Intel iGPU had issues with D3D12.

image


sp00nznet:
Landed in master via #76. 🎉

To be precise about what happened, since GitHub won't show this one as "merged": **your commits are on master under your own authorship** — the branch was folded into `draw-engine`, so the SHAs were rewritten and GitHub can't match the tip. Nothing here was re-authored or reimplemented.

The 2D/dbgfont pipeline is in, as your commits:

| on master | commit |
|---|---|
| `fa4e6b1` | rsx: 2D/dbgfont rendering pipeline + vertex-program capture |
| `a7ea577` | rsx: execute the real vertex program -> pixel-perfect dbgfont text |
| `faedc6a` | rsx: bigger vertex buffer, vsync throttle, per-clear frame reset |
| `a119f33` | cellGcmSys: recycle the FIFO ring on wrap (fix ~200-frame wedge) |
| `1df8675` | rsx: remove hardcoded debug BMP dump path |

…plus the rest of the draw engine that grew out of this branch — guest fragment-program pipelines, per-draw VP constants, indexed draws, render-to-texture, MRT, FP predication, the VS cache. 60 commits authored `sagemono <sewshee@tuta.io>` are on master now.

Closing as **merged-via-#76**, not rejected — this branch is where the renderer started.

One note for the record: #76 kept **this branch's GCM control block** (`GCM_CONTROL_GUEST_ADDR`) rather than master's `s_control_ea`, because it's what YDKJ actually renders on. Verified by booting it — 2342 flip requests in 40s, alternating 1171/1171, vs 2346 on a control built from `draw-engine` itself. No regression.

Full write-up: #76.

PR #45, opened 2026-07-05
cellNetCtl: write CellNetCtlInfo integer out-params big-endian

canersaka:
cellNetCtlGetState already byte-swaps its out-param via vm_write32, but cellNetCtlGetInfo writes the CellNetCtlInfo union's integer fields (device, mtu, link, link_type, upnp_config) through a raw GUEST_PTR struct cast with no swap, so a big-endian guest reads them byte-reversed (mtu 1500, 0x000005DC, comes back as 0xDC050000). cellNetCtlGetNatInfo has the same miss for CellNetCtlNatInfo's size and nat_type fields.

Byte-swap all of these with ps3_bswap32. The string/byte-array fields (ether_addr, ip_address, netmask, DNS) are endian-invariant, and fields set to the literal 0 are unchanged either way, so both are left as-is. Same class as the merged cellFs/cellVideoOut fixes.

Verified: the ps3recomp_runtime CMake target builds clean (MSVC, Ninja), cellNetCtl.c compiles with no new warnings.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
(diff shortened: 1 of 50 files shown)


sp00nznet:
Incorporated into `master` in [`7b6722e`](https://github.com/sp00nznet/ps3recomp/commit/7b6722e) as part of the **v0.6.6 "Two Ports, One Toolkit"** batch — reviewed, applied with your authorship preserved, credited in the README changelog + CONTRIBUTORS.md, and the runtime library builds clean. Thanks! Closing as landed.

PR #46, opened 2026-07-05
ppu_disasm: fix fsqrt/fsqrts source-register decode and print vspltis as a signed immediate

canersaka:
The A-form fsqrt/fsqrt./fsqrts handlers only special-cased XO 26/24 (the frsqrte/fres family) to print frd, frb; XO 22 (fsqrt/fsqrts) fell through to the default frd, fra, frb form, which decodes the reserved frA field as if it were a real source register (the ISA defines fsqrt's frA bits as reserved). Every fsqrt/fsqrts in a binary was therefore printed and lifted with a bogus extra operand; we measured 1258 affected sites in a retail PPU binary.

Separately, vspltisb/h/w routed their vA field through the generic vD, vA, vB register form, printing a 5-bit value the ISA defines as a signed immediate (-16..15) as an unsigned register index. They now go through a dedicated SIMM handler that sign-extends the field before printing.

Verified: py -3 tools\test_ppu_lift.py passes (1163 checks, 0 failed), and hand-encoded fsqrt/fsqrt./fsqrts and vspltisb/h/w words decode correctly across the signed-immediate range boundaries (-16, -1, 0, 15).

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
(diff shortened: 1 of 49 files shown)


sp00nznet:
Incorporated into `master` in [`127f87f`](https://github.com/sp00nznet/ps3recomp/commit/127f87f) as part of the **v0.6.6 "Two Ports, One Toolkit"** batch — reviewed, applied with your authorship preserved, credited in the README changelog + CONTRIBUTORS.md, and the runtime library builds clean. Thanks! Closing as landed.

PR #47, opened 2026-07-06
rsx: harden the NV40 VP/FP decompilers against retail-game shaders

canersaka:
Ported the NV40 VP/FP decompiler hardening from my downstream branch, found by running Yakuza: Dead Souls' real vertex and fragment programs (170 .vp / 540 .fp, captured via RPCS3's shader cache) through the decompilers and fixing every semantic gap that surfaced. The existing decompilers handled the common cases but diverged on real retail shaders in ways that produce wrong pixels rather than compile errors, so these are mostly silent-correctness fixes.

Fragment decompiler: fixed a destination store that truncated non-leading write masks (dst.zw = rhs picked rhs.xy in HLSL instead of rhs.zw).

Vertex decompiler: same component-wise store fix, plus VEC ADD reading the wrong source pair, the dest-write rule not modeling simultaneous output+temp writes and the hardware's "no write" sentinels, the scalar unit reading the wrong source field, ARL/LIT/SFL/STR falling into a generic passthrough instead of their real semantics, missing epsilon guards on RCC/RSQ/LG2, unmodeled condition-code and flow-control ops silently guessing instead of being marked, missing relative constant addressing (used for skinning matrices), only 2 of 16 vertex attributes being wired up, and HPOS bypassing the RSX viewport transform with an off-by-one output register map.

Compatibility with #43: that PR also touches rsx_vp_decompiler.c and independently fixes the same reg_type case-number swap and the same component-wise write-mask bug, at the same baseline lines. This change supersedes both fixes with a broader one covering the VEC and SCA paths, on top of the other fixes listed above. It does not touch rsx_fp_decompiler.c, so there is no overlap there, and none of #43's other files are touched by this change, so the two should merge cleanly in either order once the shared vp_decompiler hunks are reconciled.

Verified: libs/video/tests/test_fp_decompiler.c passes 9/9 (MSVC /std:c17); validate_vp_shaders.c against the Yakuza shader cache decodes 170/170 and D3DCompiles 170/170; validate_fp_shaders.c decodes 540/540 and D3DCompiles 539/540, where the one failure is an inline NaN constant literal that reproduces identically on unmodified master, so it predates this change. These prove HLSL validity and opcode coverage on the corpus, not pixel-level equivalence to hardware; a numerical diff against RPCS3 is a sensible follow-up.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
+`EDEADLK`/`EPERM` lv2 semantics (#25), and the `crnor`/`crnand` opcode-33/225
+disassembler fix (#40).
+
+### sagemono — [@sagemono](https://github.com/sagemono)
+Real-controller correctness surfaced by a DualShock-as-XInput bring-up
+(**v0.6.5**):
+- **cellPad DIGITAL2 packing** — `hs->buttons` carries DIGITAL1 in its low byte
+  and DIGITAL2 (the face buttons cross/circle/triangle/square + L1/L2/R1/R2) in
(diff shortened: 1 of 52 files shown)


sp00nznet:
Incorporated into `master` in [`8345861`](https://github.com/sp00nznet/ps3recomp/commit/8345861) as part of the **v0.6.7 "Fine Print"** batch — reviewed, applied with your authorship, credited in the changelog + CONTRIBUTORS.md, and the runtime library builds clean. Thanks! Closing as landed.

PR #49, opened 2026-07-06
spu_disasm: fix ri7 sign extension, hbr field layout, bisl/brsl link register, stop code, and channel operand width

canersaka:
Ports five SPU operand-decode fixes onto the v0.6.4 opcode-table rebuild. These are decode bugs that also feed operand extraction, so wrong operands become silently wrong lifted code rather than a visible error.

- ri7 immediates were never sign-extended, so roti with raw field 64 decoded as 64 instead of -64. Unsigned ri7 classes (cbd and friends) are unaffected and stay unsigned.
- hbr/hbra/hbrr fell through the generic RI18/RR forms; their real split branch-target and hint-address field layout is now decoded. Example: hbra word 0x10111213 at 0x45ac decodes to 0x45F8, 0x8890 instead of $r19, 0x2224.
- bisl/brsl dropped the link register operand entirely (brsl printed only the target).
- stop printed no operand; the 14-bit stop code is now shown (stop 0x102), which matters because SPURS and game kernels encode meaning in it.
- channel numbers were masked to 5 bits, wrapping channel 66 around to SPU_WrEventAck; the full 7-bit field is decoded now.

The RRR destination-register fix from the same downstream patch is deliberately not included, the v0.6.4 table rebuild already fixed that one independently.

Verified: py_compile clean, and one hand-assembled word per class diffed against the pre-fix decode output, including the hbra/hbrr oracle values from the original debugging session on real SPURS kernel code. No SPU disassembler operand suite exists upstream to run against; happy to add one as a follow-up if wanted.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
+`EDEADLK`/`EPERM` lv2 semantics (#25), and the `crnor`/`crnand` opcode-33/225
+disassembler fix (#40).
+
+### sagemono — [@sagemono](https://github.com/sagemono)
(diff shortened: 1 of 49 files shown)


sp00nznet:
Incorporated into `master` in [`28ce60c`](https://github.com/sp00nznet/ps3recomp/commit/28ce60c) as part of the **v0.6.7 "Fine Print"** batch — reviewed, applied with your authorship, credited in the changelog + CONTRIBUTORS.md, and the runtime library builds clean. Thanks! Closing as landed.

PR #50, opened 2026-07-06
lv2: extend the sub-millisecond timeout fix to event flags, semaphores, and mutexes

canersaka:
v0.6.5 fixed the sub-millisecond timeout floor in sys_event_queue_receive and headlined how much it costs, a per-frame micro-poll inflated up to 500x by the Windows timer tick. The same bug class was still live at the sibling wait sites, so any title that polls an event flag, a semaphore, or a timed mutex with a sub-ms timeout still pays the same tax there. This covers them.

sys_event_flag_wait, sys_semaphore_wait, and sys_mutex_lock now poll their wake condition to a QueryPerformanceCounter deadline (yielding with SwitchToThread between checks) instead of flooring the guest timeout to a 1 ms host wait that Windows rounds up to the ~15.6 ms tick. Each conversion is predicate-safe: the event flag pattern only mutates under the flag's lock, a semaphore post between polls stays counted in the kernel object, and TryEnterCriticalSection success is itself the atomic acquire. The QPC deadline math lives in shared helpers (lv2_usec_deadline / lv2_deadline_passed) in sys_timer.c, reused by every converted site.

Two sites deliberately not converted, for correctness rather than laziness: sys_cond_wait stays floored because this tree's sys_cond_info is a raw CONDITION_VARIABLE with no pending-signal counter, so a signal landing between two polls would be silently lost, not just delayed; fixing that needs a pending-counter rendezvous first, a separate follow-up. And sys_lwmutex_lock has no host wait loop at all here, so the class does not apply.

Verified: cl /c /std:c17 on all five touched files, zero errors, zero new warnings, checked at /W4 as well. The same conversion in my downstream tree passes its golden-boot regression with an identical milestone vector.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
+`EDEADLK`/`EPERM` lv2 semantics (#25), and the `crnor`/`crnand` opcode-33/225
+disassembler fix (#40).
+
+### sagemono — [@sagemono](https://github.com/sagemono)
+Real-controller correctness surfaced by a DualShock-as-XInput bring-up
+(**v0.6.5**):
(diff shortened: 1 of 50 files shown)


sp00nznet:
Incorporated into `master` in [`9fdedc6`](https://github.com/sp00nznet/ps3recomp/commit/9fdedc6) as part of the **v0.6.7 "Fine Print"** batch — reviewed, applied with your authorship, credited in the changelog + CONTRIBUTORS.md, and the runtime library builds clean. Thanks! Closing as landed.

PR #51, opened 2026-07-06
spu: single-round fma/fms/fnms, fesd/frds from the left word slot, and fi/frest/frsqest base+step interpolation

canersaka:
Three SPU floating-point semantic fixes, all silent-wrong-value classes found by fuzzing the lifter against per-family reference oracles and confirmed against real SPURS kernel code.

fma/fms/fnms now use fmaf() for true single rounding. The SPU's multiply-add is exact with one final rounding (CBEA p.208/210/212); plain a*b+c rounds the product first and then the sum, which diverges from hardware by a ULP on real inputs (verified with a concrete triple where the two disagree). fms becomes fmaf(a,b,-c) and fnms becomes fmaf(a,-b,c) so the single rounding covers the whole expression.

fesd/frds now read and write the left word slot of each doubleword pair (CBEA p.224/225). They previously used the right slot, so every single-to-double and double-to-single conversion operated on the wrong lane.

fi/frest/frsqest now implement the real base+step interpolation encoding (sign, biased exponent, 13-bit base fraction, 10-bit step fraction, per the public SPU ISA doc pages 214-217) instead of a plain reciprocal approximation, so the frest/fi and frsqest/fi refinement sequences produce the hardware's answers instead of results about 2x off.

One provenance note on that last part, stated openly so you can make the call: the fraction lookup tables are the SPU estimate ROM's numeric contents, hardware facts, and my transcription of them came via RPCS3's SPUThread.cpp, which itself transcribed them from hardware. The surrounding logic is written fresh against the public ISA doc, and the code comment cites the tables' nature and source. My read is that hardware ROM constants are facts and carry no license, the same numbers exist in any correct implementation, but it is your tree; if you would rather not take transcribed constants, say the word and I will split this PR so the fma and fesd/frds fixes land alone.

Verified: the header compiles standalone under cl /std:c17; a numeric harness proves fmaf(a,b,c) != a*b+c on a real triple, proves fesd/frds use the left slot with sentinel values per lane, and shows fi(x, frest(x)) tracking 1/x to about 1e-4 relative error with frsqest's sign bit forced clear per the ISA.

### DIFF
--- a/runtime/spu/spu_helpers.h
+++ b/runtime/spu/spu_helpers.h
@@ -86,8 +86,70 @@ static inline u128 spu_mpy(u128 a, u128 b)  { u128 r; for(int i=0;i<4;i++) r._s3
 static inline u128 spu_mpya(u128 a, u128 b, u128 c) { u128 r; for(int i=0;i<4;i++) r._s32[i]=(int32_t)a._s16[i*2]*(int32_t)b._s16[i*2]+c._s32[i]; return r; }
 /* sfx: extended subtract rb-ra-1+carry; carry-in = low bit of old rt (RT is 3rd src). */
 static inline u128 spu_sfx(u128 a, u128 b, u128 t) { u128 r; for(int i=0;i<4;i++) r._u32[i]=b._u32[i]+~a._u32[i]+(t._u32[i]&1u); return r; }
-/* fi: floating interpolate (reciprocal refine after frest) — Newton-step approximation. */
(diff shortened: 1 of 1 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `8d78311`:

> spu: single-round fma/fms/fnms, fesd/frds from the left word slot, and fi/frest/frsqest base+step interpolation

```
8d78311  Caner Saka <c33.saka@gmail.com>  2026-07-05
```

PR #52, opened 2026-07-06
ppu_lifter: add addeo/subfeo/mulhwo/mulhwuo overflow forms and vupklsb/vupkhsb/vupklsh unpacks

canersaka:
The lifter had no handlers for the addeo/subfeo/mulhwo/mulhwuo overflow-form mnemonics, so any occurrence fell through to the unhandled-opcode TODO stub, a silent miscompile of that instruction into a no-op. Added them following this file's existing OE-form convention (addc/subfc, addme/subfme/subfze): identical result and XER[CA] behavior as the plain form, with no XER[OV]/[SO] tracking, since none of this lifter's OE-form ops write OV. mulhwo/mulhwuo have no architected OE semantics in PowerISA at all, they are reserved encodings the shared decode can still emit, so they lift as their plain forms.

vupkhsh existed but its three siblings (vupklsh, vupkhsb, vupklsb) were decoded but unhandled. Added them mirroring vupkhsh's exact lane conventions.

Also added conformance cases for all seven ops to tools/test_ppu_lift.py, including a small vector-register case path (a vx_form encoder and a check_vr helper), since the suite previously had no VMX test support at all. That path should make future vector handler work cheaper to verify.

Verified: py_compile clean, and py -3 tools/test_ppu_lift.py passes 1295 checks, 0 failed, 0 skipped, including the new cases.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
+`EDEADLK`/`EPERM` lv2 semantics (#25), and the `crnor`/`crnand` opcode-33/225
+disassembler fix (#40).
+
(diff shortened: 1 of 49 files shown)


sp00nznet:
Incorporated into `master` in [`3d62285`](https://github.com/sp00nznet/ps3recomp/commit/3d62285) as part of the **v0.6.6 "Two Ports, One Toolkit"** batch — reviewed, applied with your authorship preserved, credited in the README changelog + CONTRIBUTORS.md, and the runtime library builds clean. Thanks! Closing as landed.

PR #53, opened 2026-07-06
tools,docs: static firmware LLE, lift a decrypted PRX and bind a game's imports to the real module

canersaka:
HLE SPURS has been the chronic wall of this toolkit, basically every SPURS title hits it, even the minimal ones. This adds the alternative that got my port past it: instead of reimplementing Sony's scheduler by hand and chasing every corner it diverges in, lift the real SPURS kernel out of a decrypted firmware PRX and run it on the recompiled runtime. The game's tasks then launch the same way they do on hardware because they are talking to the real kernel. The technique is not SPURS-specific, the same pipeline applies to any firmware module that is painful to HLE.

tools/lift_prx.py takes a decrypted PRX (ELF64 big-endian, type 0xFFA4), places its LOAD segments at a chosen guest base, applies the SCE_PPURELA relocations (ADDR32, ADDR16_LO/HI/HA, REL24, REL64, ADDR16_LO_DS, and an unrecognized type raises loudly instead of silently mis-relocating), and emits four artifacts: the relocated flat image, function seeds for ppu_lifter's existing raw-image mode (including OPD-only entry points that branch-following discovery would miss), an exports map (NID to relocated address, what a game's imports get bound to), and the module's own imports map (stub slots a runner patches to its HLE bridge). From there the existing lifter and runtime machinery do what they already do for game code.

docs/FIRMWARE_LLE.md documents the method end to end, with the prerequisite stated up front: the user supplies their own decrypted dev_flash module, this project ships no firmware and no output derived from it, and nothing of that kind is in this PR (the diff is exactly one tool and one doc). The per-game binding step is documented as prose against the existing runner pattern rather than shipped as generic code, same as the toolkit's per-game host_main convention. The doc also has a legal and redistribution section covering why the bring-your-own-firmware model keeps the tooling clean and why redistributing built binaries that embed lifted firmware is the user's own call to make.

Proven so far: on Yakuza: Dead Souls the lifted libsre SPURS kernel runs stably on the recompiled PPU/SPU runtime, the scheduler loop dispatches the game's real tasks, and the SPU-side context switch machinery is the real firmware code executing. Expected but only validated on that one title: that this unblocks other SPURS-stuck titles the same way. Fair warning from experience, the lifted kernel is real SPU code and it will surface SPU lifter bugs immediately, which is exactly why the SPU decode and semantics fixes in the other open PRs matter to it.

Verified: py_compile and --help clean on the tool, and the branch contains no game or firmware binaries of any kind.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
+`EDEADLK`/`EPERM` lv2 semantics (#25), and the `crnor`/`crnand` opcode-33/225
+disassembler fix (#40).
+
+### sagemono — [@sagemono](https://github.com/sagemono)
+Real-controller correctness surfaced by a DualShock-as-XInput bring-up
+(**v0.6.5**):
+- **cellPad DIGITAL2 packing** — `hs->buttons` carries DIGITAL1 in its low byte
+  and DIGITAL2 (the face buttons cross/circle/triangle/square + L1/L2/R1/R2) in
+  the high byte; the whole value was being written into DIGITAL1 with DIGITAL2
+  forced to 0, so every face button was dead. Split correctly (#42).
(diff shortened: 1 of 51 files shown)


sp00nznet:
Landed on `master` as [`b774c17`](https://github.com/sp00nznet/ps3recomp/commit/b774c17), authored to you, and credited in the v0.6.5 changelog + CONTRIBUTORS.md. One note: the docstring lists REL24 (type 10) and ADDR16_LO_DS (57), but `build_image` implements 1/4/5/6/44 and raises on the rest — fine for libsre, worth aligning doc vs impl before pointing it at modules that use those types. Thanks — this is the SPURS-wall unlock. Closing as incorporated.

PR #54, opened 2026-07-06
cellAudio: deliver real period events to notify queues

canersaka:
audio_notify_event_queues() was a stub that only suppressed an unused-variable warning, so registrants of cellAudioSetNotifyEventQueue never received a single event. Games that block on sys_event_queue_receive for their audio period, which is the standard cellAudio pumping pattern, waited on a queue that never got a push.

This wires the existing per-block mixing tick to actually deliver: audio_mix_thread_func already loops once per audio period, and audio_notify_event_queues() now walks the registered queues, resolves each 64-bit ipc key to a live queue through a new sys_event_find_queue_by_key() helper in runtime/syscalls/sys_event.c (mirroring how sys_event_queue_create records the key), and pushes CELL_AUDIO_EVENT_MIX through the existing sys_event_queue_push_by_id path. Registration and removal are untouched, only the firing side and the by-key lookup are new.

Supersedes my open PR #23: upstream independently built the guest-VM buffer and big-endian parameter half of that PR, so this re-submits just the still-missing piece, the event delivery, re-authored against the current tree.

Verified: both touched files compile clean with the project's flags (cl /std:c17, plus /experimental:c11atomics for cellAudio.c per the CMakeLists requirement). The same delivery path in my downstream port wakes the game's audio thread once per mixed block.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
+`EDEADLK`/`EPERM` lv2 semantics (#25), and the `crnor`/`crnand` opcode-33/225
+disassembler fix (#40).
+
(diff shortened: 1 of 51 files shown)


sp00nznet:
Incorporated into `master` in [`eb6e9d1`](https://github.com/sp00nznet/ps3recomp/commit/eb6e9d1) as part of the **v0.6.6 "Two Ports, One Toolkit"** batch — reviewed, applied with your authorship preserved, credited in the README changelog + CONTRIBUTORS.md, and the runtime library builds clean. Thanks! Closing as landed.

PR #55, opened 2026-07-06
libs: correct HLE ABI signatures for cellPamf, cellSail, cellHttp/cellSsl, cellNet/sysNet, and cellGameData

canersaka:
Several HLE stubs had signatures that do not match the real PS3 ABI, so the generic import bridge marshalled the wrong registers into them, dropping real caller data or handing the callee garbage. These are silent failures: the stub runs, returns OK, and the corruption shows up later.

Two concrete examples. cellSailMemAllocatorInitialize took two raw function pointers plus an arg pointer as three parameters, but the guest passes one pointer to a CellSailMemAllocatorFuncs struct with two callback fields, so alloc and free each ended up holding half of a guest address instead of the two callbacks. sys_net_initialize_network_ex took (mem_container, flags) as two u32s, but the real function takes a single pointer to a sys_net_initialize_parameter struct or NULL, so the actual parameter pointer was treated as a container id and uninitialized r4 was read as flags.

Also fixed: cellPamfReaderGetMuxRateBound and cellSailPlayerSetRepeatMode return their result directly in r3 rather than via an out pointer, cellHttpInit, cellHttpCreateTransaction, cellHttpsInit/End, cellHttpUtilUnescapeUri, and cellSslCertificateLoader had extra, missing, or reordered parameters relative to hardware, cellNetCtlTerm returns void not s32, and cellGameCreateGameData took a CellGameContentSize*/dirName pair instead of the real CellGameSetInitParams* plus two output path buffers, so the game's init struct was being read as an output size struct.

sceNpTrophyRegisterContext and cellNetCtlGetInfo are deliberately untouched, the former was already corrected upstream with real callback semantics, and the latter's byte-swap issue is #45.

Verified: full ps3recomp_runtime static library builds and links clean (CMake, Ninja, MSVC) with the changes, no new warnings. Found by an ABI audit tool that cross-checks stub signatures against the dispatch classifier, which is part of the audit-kit PR coming separately.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
+`EDEADLK`/`EPERM` lv2 semantics (#25), and the `crnor`/`crnand` opcode-33/225
+disassembler fix (#40).
+
+### sagemono — [@sagemono](https://github.com/sagemono)
+Real-controller correctness surfaced by a DualShock-as-XInput bring-up
+(**v0.6.5**):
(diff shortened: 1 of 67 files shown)


sp00nznet:
Incorporated into `master` in [`a7e3ecd`](https://github.com/sp00nznet/ps3recomp/commit/a7e3ecd) as part of the **v0.6.7 "Fine Print"** batch — reviewed, applied with your authorship, credited in the changelog + CONTRIBUTORS.md, and the runtime library builds clean. Thanks! Closing as landed.

PR #56, opened 2026-07-06
tools: add a lifter/HLE audit kit (decode cross-checks, operand audits, HLE ABI audit, endianness heuristics, lift-structure census) and an lv2 sync stress test

canersaka:
This ships the differential audit tooling I used to find and confirm the lifter and HLE bugs that several of the other open PRs fix, generalized for any PPU ELF, HLE tree, or import-bridge convention rather than hardcoded to my port. The reason it exists: a decode-level swap, two opcodes trading mnemonics, survives every runtime and conformance test in the tree, because the lift stays internally self-consistent and only disagrees with the real architecture. A self-referential test suite cannot catch that, only an external oracle can. The crnor/crnand opcode swap, the SPU operand decode fixes, and the HLE ABI signature corrections in the companion PR were all found this way, so this is the reusable machinery behind that work, not a one-off.

disasm_audit.py and disasm_audit_operands.py cross-check tools/ppu_disasm.py against Capstone over an ELF's executable segments, first at the mnemonic level, then at the operand level for the words where the mnemonics already agree. Both carry seeded self-tests that deliberately introduce a swapped-opcode bug and confirm the harness flags it, so a clean run means the checker is actually working rather than silently passing.

spu_disasm_audit.py and spu_disasm_audit_operands.py do the same for tools/spu_disasm.py against RPCS3's SPUDisAsm. RPCS3 is used strictly as an output-only oracle you build yourself against your own RPCS3 clone and diff text against, so no GPL source is ever copied into this repo, the same clean-room posture the rest of the toolkit keeps.

endian_audit.py is an advisory heuristic scan of the HLE C sources for out-params that look like they may be missing a big-endian byte swap. hle_abi_audit.py diffs our HLE function signatures against RPCS3's real module implementations, and hle_dispatch_classify.py layers on top of it to say which of those mismatches are actually live at runtime versus shadowed by an LLE-dispatched or override-wrapped import slot, given a generated NID-to-function-pointer bridge table, so you triage the ones that matter instead of chasing dead stubs. lift_audit.py censuses a relift's structural health: unresolved lift markers by mnemonic, unresolved branch targets, and fallthrough-into-the-next-function hazards, with an optional baseline diff so regressions across relifts show up as a diff instead of a wall of text.

tests/sync_stress is a standalone lv2 sync-primitive stress suite (mutex, cond, semaphore, event queue) that links the real runtime sources directly and drives them from many host threads, to catch lost wakeups, timed-wait regressions, counting-invariant violations, and event-delivery ordering bugs under real concurrency, independent of the full game runner. Stated openly: it currently fails 2 of 5 cases against this repo's checked-in runtime, both real and narrow, a semaphore post that returns CELL_EINVAL instead of CELL_EBUSY when the post would exceed max_value, and an event-queue receive that never writes the dequeued event into gpr4 through gpr7 so it is silently lost. Both are exactly the class of bug this kit exists to surface, so I left them visible rather than whitelisting them away. Each whitelist file suppresses only known intentional divergences, so a clean run means zero unexpected findings.

Verified: all eight Python tools compile clean under Python 3.14 (py -3 -m py_compile) and are stdlib-only. The disasm cross-checks require an installed capstone and the SPU oracle requires a locally built RPCS3, both external to this repo and documented in each script's header. Not exercised here: a full corpus-wide audit run, since that needs the external oracles set up; the findings above came from running the kit against my own port's tree.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -62,6 +62,22 @@ wrong. Landed in **v0.6.1 "Many Hands"**:
 - **`tools/show_func.py`** — dump a single lifted function's C and/or its
   original PowerPC disassembly from the chunked output.
 
+*Also incorporated (**v0.6.5**)* — cellFs big-endian out-params + `CellFsStat`
+PS3 packing (#22), cellGame title id read from `PARAM.SFO` (#24), `sys_rwlock`
+`EDEADLK`/`EPERM` lv2 semantics (#25), and the `crnor`/`crnand` opcode-33/225
+disassembler fix (#40).
+
+### sagemono — [@sagemono](https://github.com/sagemono)
+Real-controller correctness surfaced by a DualShock-as-XInput bring-up
+(**v0.6.5**):
+- **cellPad DIGITAL2 packing** — `hs->buttons` carries DIGITAL1 in its low byte
+  and DIGITAL2 (the face buttons cross/circle/triangle/square + L1/L2/R1/R2) in
+  the high byte; the whole value was being written into DIGITAL1 with DIGITAL2
+  forced to 0, so every face button was dead. Split correctly (#42).
+- **analog-Y centering** — reflect the inverted Y about 128 (`256 - x`) instead
+  of `255 - x`, which turned a centered stick into 127 (`0x7F`, aliasing
+  SELECT+START self-exit + TRIANGLE).
+
 ### Paulo Adriano Alves — [@pauloadrianoalves](https://github.com/pauloadrianoalves)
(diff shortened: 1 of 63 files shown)


sp00nznet:
Incorporated into `master` in [`4a3cea6`](https://github.com/sp00nznet/ps3recomp/commit/4a3cea6) as part of the **v0.6.7 "Fine Print"** batch — reviewed, applied with your authorship, credited in the changelog + CONTRIBUTORS.md, and the runtime library builds clean. Thanks! Closing as landed.

PR #57, opened 2026-07-08
lv2: implement sys_spu_image_import & fix SPU image syscall numbers

sagemono:
Start initial work on SPU execution.

### DIFF
--- a/runtime/syscalls/lv2_register.c
+++ b/runtime/syscalls/lv2_register.c
@@ -952,20 +952,97 @@ uint8_t* spu_thread_get_local_store(uint32_t tid)
 
 uint32_t spu_thread_local_store_size(void) { return SPU_LS_SIZE; }
 
-/* sys_spu_image_import(*img, *source, type) — just log entry & return success.
- * We could parse the SPU ELF header and write entry into the image struct,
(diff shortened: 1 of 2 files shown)


sp00nznet:
Landed in master via #76. 🎉

Same story as #43 — GitHub won't mark this "merged" because the branch was folded into `draw-engine` and the SHAs were rewritten, but **your implementation is on master under your own authorship**, as `bcf2c1b` (`sagemono <sewshee@tuta.io>`):

> lv2: implement sys_spu_image_import; fix SPU image syscall numbers

Both halves are live on master:

```
#define SYS_SPU_IMAGE_IMPORT  157   /* _sys_spu_image_import (RPCS3 lv2.cpp) */
#define SYS_SPU_IMAGE_CLOSE   158   /* was mis-numbered 157 -> import hit the close stub */
```

…and the SPU ELF parse (entry point + segment table, `PT_LOAD` → COPY, BSS tail → FILL). That's the version #76 kept: master had only a log-and-return-success stub, and `research/proto-builds` had an entry-point-only parse with no segments — yours was the complete one, so it won.

#63 builds on this with the `_sys_spu_image_import` wrapper (`0xEBE5F72F`) and already credits you; it stays open on its own merits.

PR #58, opened 2026-07-08
spu_lifter: brsl target parse broken by the disasm link register fix, every call lifted as a silent no-op

canersaka:
This is a miscompile class bug that is live on master right now, so flagging it with some urgency: since the disasm fix that added the link register to brsl and brasl operand text (brsl $r0, 0x5488), spu_lifter's _branch_target still reads ops[0], gets the register token instead of the target, and silently returns None. Every brsl in the module then hits the TODO fallback and is emitted as a no-op that only sets the link register. The module compiles and runs, but every call inside it is skipped. The comment in spu_disasm.py saying lift output is unaffected covers the link register recovery only, not the target parse.

The failure is invisible at lift time except for one line, 0 call target(s) in the summary. A SPURS job binary lifted with the broken pair ran straight through its body without issuing its DMA, its event flag write, or its interrupt mailbox doorbell, and everything downstream of it looked like a scheduling bug. Any SPU module lifted since the disasm change has this, so anything regenerated since then should be relifted after this lands.

The fix takes the target from the last operand for the link forms, which handles both the old and new operand text, and the lift summary now warns loudly with addresses whenever a brsl or brasl target fails to resolve, so a dropped call can never be silent again.

Verified: on this repo's current tree, lifting the same SPURS job binary gives 0 call target(s) unfixed and 10 with this fix, matching a lift made before the disasm change; at runtime the module went from returning without doing its work to issuing its calls.

### DIFF
--- a/tools/spu_lifter.py
+++ b/tools/spu_lifter.py
@@ -639,14 +639,20 @@ class SPULifter:
             return ("ctx->pc = ctx->srr0; "
                     "spu_indirect_branch(ctx); return;")
         if mn in ("bisl",):
-            link_rt = insn.raw & 0x7F            # link reg dropped from operands
-            tgt_reg = _reg(ops[0])
+            link_rt = insn.raw & 0x7F
+            # Same interface break as brsl: the disasm fix added the link
+            # register to bisl operand text ("bisl $r0, $r11"), so the TARGET
+            # register is now the LAST operand. ops[0] grabbed the LINK
+            # register, so the emitted indirect call dispatched to link
+            # (= addr+4) and the callee was silently skipped. ops[-1]
+            # handles both operand formats.
+            tgt_reg = _reg(ops[-1])
             return (f"{g(link_rt)} = spu_link(0x{addr + 4:X}); "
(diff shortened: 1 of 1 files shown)


canersaka:
Pushed a second commit: bisl and bisled have the identical break from the same disasm operand change (their text also gained the link register, so the ops[0] parse grabs link instead of the target and the indirect call silently falls through to the next instruction). Found it by sweeping the whole class after the brsl fix: a raw encoding vs lifter output cross check over every branch and call in my corpus plus a full EBOOT, 567787 instructions. Happy to PR that audit tool separately if you want it, it exits nonzero on any disagreement so it can gate lifts the way a conformance suite does.

sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `4f98318`:

> spu_lifter: bisl and bisled target register is the last operand, same class as brsl

```
4f98318  Caner Saka <c33.saka@gmail.com>  2026-07-08
```

PR #60, opened 2026-07-08
ppu: FP correctness batch (fcmpu NaN, fcti saturation, single rounding, vctsxs/vctuxs)

canersaka:
Four FP correctness classes found while booting Yakuza through the lifter, all cited against PowerISA Book I and validated on my side. fcmpu and fcmpo now report FU (0b0001, unordered) when either operand is NaN per 4.6.8, where the old emission had no NaN branch at all and took a comparison arm on NaN poisoned data. fctiw, fctiwz, fctid and fctidz now saturate (positive overflow to max, negative and NaN to min) instead of a raw C cast that is undefined behavior on overflow. The ten single precision forms (fadds family, frsp, fsqrts, the fused fmadds family) now round their results to float precision instead of silently keeping full double precision, which changes downstream comparisons in real code. vctsxs and vctuxs saturate the same way. The VMX changes keep this repo's native endian cast convention rather than importing mine.

Verified: 1379 of 1379 conformance checks green on this tree, including 256 new reference cases covering all four classes at cr0 and cr7 and across the overflow, NaN and exact boundary inputs; reverting the lifter change alone and rerunning makes 41 of the new cases fail across all four classes, so the suite discriminates old from new behavior rather than passing vacuously.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -116,6 +116,28 @@ SOURCE_PREAMBLE = """\
 #include <string.h>
 #include <math.h>
 
+/* Float->int conversion per Book I 4.6.7: SATURATE out-of-range (positive
+ * overflow => max, negative => min, NaN => min) and honor the rounding mode
+ * (rn=1: FPSCR default round-to-nearest-even via nearbyint; rn=0: the z
+ * forms truncate). A plain C cast here is UB on overflow/NaN and MSVC
+ * sign-flips positive overflow to INT_MIN on x86. */
+static inline uint32_t ppu_f2i32(double v, int rn)
+{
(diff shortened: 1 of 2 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `b9761b1`:

> ppu: FP correctness batch (fcmpu NaN, fcti saturation, single rounding, vctsxs/vctuxs)

```
b9761b1  Caner Saka <c33.saka@gmail.com>  2026-07-08
```

PR #61, opened 2026-07-08
ppu: decode and lift stvlx/stvrx/stvlxl/stvrxl (Cell unaligned vector stores)

canersaka:
stvlx, stvrx and their cache hint forms were absent from both the disassembler and the lifter, so a title using Cell unaligned vector stores compiled to silent memory non writes and the corruption surfaced far from the store site. This adds decode and lift for all four, using the same raw byte convention as the existing lvlx and lvrx loads and this repo''s _xea helper style.

Verified: 1304 of 1304 conformance checks green on this tree including 117 new cases across aligned, every unaligned offset, and the zero length edge (EA aligned means stvrx stores nothing); a smoke lift of all four mnemonics emits real byte copy loops, not a TODO fallback.

### DIFF
--- a/tools/ppu_disasm.py
+++ b/tools/ppu_disasm.py
@@ -790,6 +790,27 @@ def decode(insn: int, addr: int = 0) -> Instruction:
             result.operands = f"r{rd}, r{ra}, r{rb}"
             return result
 
+        # Cell unaligned vector STORES (were previously undecoded, so any
+        # vector store using them fell through to the op31_x fallback and got
+        # lifted as a TODO no-op -- a silent memory non-write, not just a
+        # wrong value).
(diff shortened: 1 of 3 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `a37a029`:

> ppu: decode and lift stvlx/stvrx/stvlxl/stvrxl (Cell unaligned vector stores)

```
a37a029  Caner Saka <c33.saka@gmail.com>  2026-07-08
```

PR #62, opened 2026-07-08
ppu: value verified CAS for lwarx/stwcx and ldarx/stdcx

canersaka:
stwcx and stdcx succeeded on an address only reservation check and then did a plain store. Under contention that silently reverts a concurrent writer: the other thread''s value lands between the lwarx and the stwcx, the store overwrites it anyway, and the loser''s update just vanishes with nothing to see in a log. Guest lock free structures (SPURS management lines are built on exactly this pattern) corrupt rarely and unreproducibly. The store conditional is now a real compare and swap on the reserved value through atomic builtins, with both the MSVC and clang paths covered, and a failed CAS clears the reservation and reports failure to CR0 the way hardware does.

One note for review: runtime/ppu/ppu_memory.h already contains a value correct lwarx/stwcx pair, but it is dead code, zero includers anywhere in the tree, with a signature incompatible with the lifter''s emission. This PR adds self contained helpers instead and flags that module in the commit message rather than reviving it.

Verified: 1311 of 1311 conformance checks green on this tree including the exact race the fix targets, address matches but the value was changed by a simulated concurrent writer between the load and the store conditional, which must fail and preserve the writer''s value, and does.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -168,6 +168,87 @@ static inline uint64_t ppc_mulhdu(uint64_t a, uint64_t b) {
 /* VM base pointer (defined by game project) */
 extern "C" uint8_t* vm_base;
 
+/* PPU reservation atomics (lwarx/stwcx/ldarx/stdcx): value-verified CAS.
+ * PowerISA Book II 4.6.2/4.6.3: stwcx./stdcx. succeed only if the reservation
+ * granule still holds the value observed at the matching lwarx/ldarx -- any
+ * store to that granule by another processor between the two must be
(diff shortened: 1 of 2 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `4bfc107`:

> ppu: value verified CAS for lwarx/stwcx and ldarx/stdcx

```
4bfc107  Caner Saka <c33.saka@gmail.com>  2026-07-08
```

PR #66, opened 2026-07-09
lv2: sys_spu_thread_group_create reads r5 as priority, name from the attr struct

canersaka:
sys_spu_thread_group_create_handler in runtime/syscalls/lv2_register.c read r5 as a guest name pointer and never captured the group priority. The real ABI, per RPCS3's sys_spu.cpp:940 (oracle, behavior only, no code copied), is sys_spu_thread_group_create(id, num, prio, attr): r5 is an s32 priority and the name lives in the attribute struct, whose layout (sys_spu.h:85-90) is nsize at offset 0, the name pointer at 4 and an s32 type at 8. A caller passing a real priority, SPURS uses values like 100 and 250, had that integer dereferenced as a guest address, so the group name came back as garbage or a stray read from low guest memory, and the actual priority was silently dropped.

The handler now reads prio from r5 and nsize, name and type from the attr struct through the file's existing vm_read_be32 helper.

Verified: ps3recomp_runtime rebuilds clean with this as the only changed file (cmake and ninja, 112 of 112 objects)

### DIFF
--- a/runtime/syscalls/lv2_register.c
+++ b/runtime/syscalls/lv2_register.c
@@ -297,15 +297,23 @@ static int64_t sys_spu_initialize_handler(ppu_context* ctx)
     return 0;
 }
 
-/* sys_spu_thread_group_create(out_id_ea, num, name_ea, attr_ea) */
+/* sys_spu_thread_group_create(out_id_ea, num, prio, attr_ea)
+ *
+ * lv2 signature per RPCS3's sys_spu.h (oracle, no code copied): r5 is the
+ * group PRIORITY (an int), NOT a name pointer. The name lives inside the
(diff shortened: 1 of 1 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `89afe3e`:

> lv2: sys_spu_thread_group_create reads r5 as priority, name from the attr struct

```
89afe3e  Caner Saka <c33.saka@gmail.com>  2026-07-08
```

PR #68, opened 2026-07-09
ppu: lmw and stmw load store multiple word

canersaka:
﻿lmw and stmw were already decoded by ppu_disasm but fell through to the lifter's TODO catch all, so any guest function using either instruction silently dropped the whole load or store instead of touching the registers or memory at all. This is a port from the Yakuza port's lifter. PowerISA_V2.03_Final_Public.pdf page 72, Book I section 3.3.5, gives EA as (RA=0 ? 0 : GPR[RA]) plus D, then each register from RT (or RS) through r31 is a plain four byte big endian word at EA, EA+4 and so on, RA=0 meaning a literal zero base rather than GPR[0], called out explicitly since the plain loads and stores just above this block do not handle rA=0 that way. These are the named lmw_last and stmw failure classes in the PR 64 torture suite.

Verified: 1303 of 1303 conformance checks green (993 cases, 0 skipped) including 4 new cases covering both directions of the transfer and both the normal and RA=0 addressing forms, backed by a small vm stub added to the conformance suite. Reverting the lifter change alone and keeping the new test cases makes all 4 of the new cases skip as unhandled by the lifter instead of running, since lmw and stmw fall back to the TODO stub without the fix, 989 cases run, 1295 checks passed, 0 failed, 4 skipped, so the suite discriminates old from new behavior.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -822,6 +822,41 @@ class PPULifter:
                 return line
             return f"/* {mn} unhandled operands: {insn.operands} */;"
 
+        # ------- Load/Store Multiple Word -------
+        # PowerISA_V2.03_Final_Public.pdf p.72 (Book I, section 3.3.5):
+        #   lmw RT,D(RA):  b = (RA=0) ? 0 : GPR[RA]; EA = b + EXTS(D);
+        #                  for r = RT..31: GPR(r) = 0x0..0 || MEM(EA,4); EA += 4
+        #     (n = 32-RT words load into the LOW-order 32 bits of GPRs RT..31,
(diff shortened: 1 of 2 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `2be0eba`:

> ppu: lmw and stmw load store multiple word

```
2be0eba  Caner Saka <c33.saka@gmail.com>  2026-07-09
```

PR #70, opened 2026-07-09
ppu: mulldo overflow form

canersaka:
﻿mulldo and mulldo. are already decoded by ppu_disasm through the same OE suffix decode already used for mulhwo, but the lifter only matched an exact mulld, so a function computing a 64 bit product and checking overflow got the OE bit silently ignored, correct low 64 bits but XER[OV] never set. This is a port from the Yakuza port. PowerISA_V2.03_Final_Public.pdf page 60 and 36 say OE=1 sets XER[OV] to 1 when the true 128 bit signed product does not fit in 64 bits, meaning the high half is not just the sign extension of the low half, and XER[SO] sticky ors with OV. The fix reuses ppc_mulhd, already emitted here for mulhd, to get that high half without signed multiply overflow undefined behavior, so no new helper is needed. This intentionally departs from how the file's other OE form ops behave, adde/subfe and mulhw/mulhwu lift the OE encoding identically to the plain form because their overflow condition is not tracked at all, but mulld's condition is directly computable from a multiply the lifter already performs, so it gets real overflow tracking instead of the file's usual no OV convention for OE forms.

Verified: 1307 of 1307 conformance checks green (995 cases, 0 skipped) including the 6 new cases, 3 value pairs, positive overflow, negative overflow, no overflow, each run through both mulldo and mulldo. Reverting the lifter change alone and keeping the new test cases makes all 6 of the new cases skip as unhandled by the lifter instead of running, since mulldo falls back to the TODO stub without the fix, 989 cases run, 1295 checks passed, 0 failed, 6 skipped, so the suite discriminates old from new behavior.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -1241,8 +1241,22 @@ class PPULifter:
                     f"ctx->cr = (ctx->cr & ~(0xFu << {shift})) | (cr_val << {shift}); }}")
 
         # ------- 64-bit multiply / divide -------
-        if mn == "mulld":
+        # mulldo[.]: PowerISA_V2.03_Final_Public.pdf p.60/p.36 -- OE=1 sets
+        # XER[OV] = 1 iff the true 128-bit signed product does not fit in 64
+        # bits (the high 64 bits aren't just the sign-extension of the low),
+        # XER[SO] sticky-ORs with OV. Uses ppc_mulhd (already emitted for
+        # mulhd, above) for the exact high half, avoiding signed-multiply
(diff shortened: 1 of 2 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `dfcefb2`:

> ppu: mulldo overflow form

```
dfcefb2  Caner Saka <c33.saka@gmail.com>  2026-07-09
```

PR #71, opened 2026-07-09
ppu: indexed store with update writes back ra

canersaka:
﻿stbux, sthux and stwux sat in the plain idx_store_map alongside the non update indexed stores, so they stored correctly but never wrote RA back to the effective address, a landmine since Book I defines store with update as RA gets EA in addition to the store itself. A dedicated, correct stwux handler existed further down the same if chain but could never run because the plain map's match comes first, dead shadowed code that looked like coverage but was not reachable. The fix moves all three mnemonics into their own idx_store_update_map that computes EA once, stores through it, then writes RA, and removes the now redundant dead handler. This class has zero live call sites in the game today, the 64 bit ABI uses stdux instead, but it is exactly the kind of silent RA corruption that is expensive to chase later, so I am removing it now rather than waiting for it to bite.

Verified: 1301 of 1301 conformance checks green (992 cases, 0 skipped) including the 3 new cases, stwux, sthux and stbux, each checking both the stored memory bytes and RA. Reverting the lifter change alone and keeping the new test cases makes 3 of the 3 new cases fail on the RA check while the memory write itself still passes, 992 cases still run, 1298 of 1301 checks passed, 3 failed, so the suite discriminates old from new behavior.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -854,13 +854,25 @@ class PPULifter:
         # ------- Indexed Stores -------
         idx_store_map = {
             "stbx": "vm_write8", "sthx": "vm_write16", "stwx": "vm_write32", "stdx": "vm_write64",
-            "stbux": "vm_write8", "sthux": "vm_write16", "stwux": "vm_write32",
         }
         if mn in idx_store_map:
             helper = idx_store_map[mn]
             rs_i, ra_i, rb_i = _reg_idx(ops[0]), _reg_idx(ops[1]), _reg_idx(ops[2])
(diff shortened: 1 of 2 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `32510fd`:

> ppu: indexed store with update writes back ra

```
32510fd  Caner Saka <c33.saka@gmail.com>  2026-07-09
```

PR #72, opened 2026-07-09
ppu: d form loads and stores treat ra 0 as a literal zero base

canersaka:
﻿I found this as a follow up while porting the Yakuza port's store update fix over. The plain D form load and store maps in the lifter, lbz through ld and stb through std, plus the DS form lwa, emitted ctx->gpr[base] + disp unconditionally and never checked whether rA was r0. The same pattern was sitting in the D form floating point loads and stores a little further down the same file, lfs, lfd, stfs and stfd, so those get the identical guard in this change too. PowerISA_V2.03_Final_Public.pdf page 64, Book I section 3.3.2, spells out the D form load rule as EA equals, if RA is 0 then b gets 0 else b gets the contents of RA, plus D, and page 67 gives the identical rule for the D form stores, both explicit that RA equal to 0 means a literal zero base, not the contents of GPR 0, and the floating point loads and stores are D form instructions under the same rule. The indexed X form loads and stores already apply this correctly a few lines further down in the same file, so the plain integer and floating point forms were the odd ones out.

This is a landmine because any absolute address idiom that encodes as disp(0) reads or writes through whatever garbage happens to sit in r0 at that point instead of the intended fixed address, and it only shows up later as corrupted memory somewhere downstream of the instruction that actually misbehaved. Update forms are correctly left alone, since the manual states that RA equal to 0 on an update form is itself an invalid instruction, so there is nothing to guard there.

Verified: 996 of 996 conformance cases green, 1302 of 1302 checks passed, 0 skipped, including the 5 integer cases covering lwz and stw with rA equal to 0 against a poisoned r0, lwz and stw with a normal rA for a no regression check, and lwa with rA equal to 0 for the DS form side, plus 2 more for the floating point side, lfd with rA equal to 0 against a poisoned r0 checked against a known double bit pattern, and stfs with rA equal to 0 checked against the stored single precision bytes. Reverting just the lifter hunk and keeping all the new cases turns 5 of them red, the lwz, stw, lwa, lfd and stfs rA equal to 0 cases each fail on the wrong address while the two rA not equal to 0 regression cases keep passing, 1297 of 1302 checks passed, 5 failed, 0 skipped, so the suite discriminates the old behavior from the fixed one.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -779,6 +779,12 @@ class PPULifter:
             return f"ctx->gpr[{ra}] = (uint64_t)ppc_srad(&ctx->xer, (int64_t)ctx->gpr[{rs}], {_imm(ops[2])});"
 
         # ------- Loads -------
+        # PowerISA_V2.03_Final_Public.pdf p.64 (Book I, section 3.3.2, "Load
+        # Word and Zero D-form"): for a D-form load, EA = (RA=0 ? 0 : GPR[RA])
+        # + D -- RA=0 is a LITERAL ZERO base, not GPR[0]. Update forms (lwzu
+        # etc.) are excluded from the rule: p.62 spells out "If RA=0 ... the
+        # instruction form is invalid" for the update forms, so no guard is
+        # needed for the base += disp below.
         load_map = {
             "lbz": ("vm_read8", False), "lbzu": ("vm_read8", False),
             "lhz": ("vm_read16", False), "lhzu": ("vm_read16", False),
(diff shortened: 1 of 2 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `8d5f31a`:

> ppu: d form loads and stores treat ra 0 as a literal zero base

```
8d5f31a  Caner Saka <c33.saka@gmail.com>  2026-07-09
```

PR #73, opened 2026-07-09
ppu: fp load and store update forms write back ra

canersaka:
﻿lfsu, lfdu, stfsu and stfdu sat in the plain lfs/lfd/stfs/stfd handlers alongside their non update counterparts, so the access happened but RA never got EA written back. PowerISA_V2.03_Final_Public.pdf p.129 gives Load Floating Point Single with Update as EA computed then RA gets EA, p.130 gives the same rule for the double form, and p.132 and p.133 give the identical rule for the single and double stores with update, all four stating that RA=0 on these forms is itself an invalid instruction, so unlike the plain D form loads and stores there is no rA=0 literal zero base case to guard here. This is the FP sibling of the indexed integer fix on the ppu-store-update-writeback branch, same landmine class, the base pointer idiom that walks a struct of floats or doubles one field at a time never advances, so it re reads or re writes the same field forever instead of stepping through memory. I checked whether the X form update variants share the defect, lfsux, lfdux, stfsux and stfdux are decoded by ppu_disasm.py but have no handler at all in ppu_lifter.py, so they fall through to the unhandled catch all rather than silently miscompiling, a separate pre existing gap noted in tools/disasm_audit_operands_whitelist.txt and left alone here since the fix is scoped to the D form defect that was actually confirmed. The fix gives each of the four D form mnemonics its own handler that computes EA once, performs the load or store through it, then writes RA, leaving the non update lfs/lfd/stfs/stfd handlers untouched.

Verified: 994 of 994 conformance cases green, 1305 of 1305 checks passed, 0 skipped, including 5 new cases, lfsu, lfdu, stfsu and stfdu each checking both the memory access and the updated RA, plus an lfd regression check confirming the non update form leaves RA untouched. Reverting just the lifter hunk and keeping all five new cases turns exactly the 4 RA checks red while the memory access side of each case keeps passing, 1301 of 1305 checks passed, 4 failed, 0 skipped, so the suite discriminates the old behavior from the fixed one.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -1105,7 +1105,7 @@ class PPULifter:
             return "lv2_syscall(ctx);"
 
         # ------- Floating-point loads/stores -------
-        if mn in ("lfs", "lfsu", "lfd", "lfdu"):
+        if mn in ("lfs", "lfd"):
             frd = _reg_idx(ops[0])
             disp, base = _disp_base(ops[1])
             if disp is not None:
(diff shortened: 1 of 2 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `9d728a9`:

> ppu: fp load and store update forms write back ra

```
9d728a9  Caner Saka <c33.saka@gmail.com>  2026-07-09
```

PR #74, opened 2026-07-09
ppu: vmx element wise ops read big endian lanes, values were byte reversed on x86

canersaka:
﻿I found this while working the Yakuza port: the VMX element wise handlers in this tree read and write vector lanes through native typed pointer casts, and on an x86 host that silently byte reverses every multi byte lane value before any math happens to it. ctx->vr holds raw big endian bytes, since lvx and stvx are plain memcpys and vm_read32 in ppu_loader.cpp proves that convention by explicitly bswapping on read, but a handler like vcmpgtsw just does (int32_t*)&ctx->vr[va] and compares, so 256 stored as big endian bytes 00 00 01 00 gets read natively as 0x00010000, which is 65536, and 2 stored as 00 00 00 02 reads as 0x02000000, which is 33554432, so the compare returns false when the real answer is 256 is greater than 2, true. The same thing happens to every add, subtract, min, max, multiply, shift, splat immediate, convert, and pack instruction that operates on halfword, word, or float lanes, and the float ops have it worse, since reinterpreting a byte reversed IEEE 754 bit pattern does not even land near the right magnitude, so vaddfp, vmaddfp, vrefp, vcfsx, and friends were computing on essentially random values. Tests missed this for a long time because every byte width op, vperm, vspltb, vand, vxor, vsldoi, and so on, is exactly invariant to the byte order confusion, so anything exercising only those looked fine, and single byte values in a multi byte lane also happen to survive a reversal unchanged, which hides the bug further.

The fix ports the swap aware lane accessor pattern I already use in the Yakuza port: vrh, vrw, vrd, and vrf do a memcpy plus a width appropriate byteswap on read, vsth, vstw, vstd, and vstf byteswap then memcpy on write, and every affected handler now routes through them, the compare family, the add and subtract family, min and max, the float arithmetic and float compares, the int and float convert instructions, the shifts and rotates, the multiplies, the sum and multiply accumulate instructions, the pack, and the unpacks. Storage itself never changes, it stays raw big endian, so lvx and stvx keep doing their plain memcpy exactly as before, only the accessors swap. Byte width lanes and pure bitwise ops are deliberately left alone, and I want to call out one counterintuitive case, vspltw and vsplth read as multi byte but turn out to already be correct, since a splat only copies an existing lane's bytes to other lane positions with no arithmetic in between, so reading and writing with the same reinterpretation cancels out, and I added a case for both to the suite specifically to document that as a control rather than silently skip them.

Verified: py -3 tools/test_ppu_lift.py passes 1301 checks, 0 failed, 0 skipped, out of 995 generated cases with 0 skipped at generation. I added conformance cases chosen to discriminate the bug rather than just exercise the instruction, vcmpgtsw and vcmpgtsh on 256 versus 2, vadduwm on values that straddle a byte boundary, and vmaddfp on exact 2.0 family values, and corrected the four vupk* reference functions in the suite that had been modeling the byte reversed output as the intended behavior. Reverting just the lifter change and rerunning against the same new cases fails exactly 16 of them, the four discriminating cases plus the 12 vupk* cases whose reference changed, while the vspltw and vsplth control cases stay green in both states as expected. Every element wise VMX compare, add, splat immediate, and float op in this file computed the wrong value for any lane wider than one byte on an x86 host before this change.

### DIFF
--- a/tools/ppu_lifter.py
+++ b/tools/ppu_lifter.py
@@ -1663,7 +1663,14 @@ class PPULifter:
                     f"tmp[i] = (sel < 16) ? a[sel] : b[sel - 16]; }} "
                     f"memcpy(d, tmp, 16); }}")
 
-        # VMX splat (vspltw, vsplth, vspltb) — duplicate one element across vector
+        # VMX splat (vspltw, vsplth, vspltb) — duplicate one element across
+        # vector. Left as native casts deliberately: a splat only ever COPIES
+        # an existing lane's raw bytes to other lane positions (no arithmetic
+        # on the value), so reading and writing with the same native
+        # reinterpretation cancels out and the transferred bytes are correct
+        # regardless of host endianness. (Unlike vspltisb/h/w two blocks
+        # above, which synthesize a NEW value from an immediate and do need
+        # the accessors.) See vsplth below for the halfword form.
         if mn == "vspltw":
             vd = int(ops[0][1:])
             vb = int(ops[1][1:])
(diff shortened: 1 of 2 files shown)


sp00nznet:
Landed in master via #76. 🎉

GitHub won't flag this one as "merged" — the branch was folded into `draw-engine` during integration, so the SHA was rewritten and the tip no longer matches. To be clear about what that means: **your change is on master under your own authorship**, as `a8ca0f4`:

> ppu: vmx element wise ops read big endian lanes, values were byte reversed on x86

```
a8ca0f4  Caner Saka <c33.saka@gmail.com>  2026-07-09
```

PR #83, opened 2026-07-25
cellAudio: clear each consumed ring buffer block

canersaka:
The audio consumer advanced the read index after mixing a block but left the consumed samples in guest memory. A producer that does not completely overwrite that slot before it wraps around can therefore replay stale PCM from the previous trip through the ring. RPCS3's cellAudio consumer clears the block before advancing the read position for the same reason.

This zeros exactly the consumed block after it has been copied into the host mix buffer and before the read index is published. The producer-visible read-index behavior and the existing sample conversion stay unchanged.

Verified: ps3recomp_runtime builds clean from the current ps3recomp staging base with this as the only changed file. I also traced the consume order against RPCS3's reset_ports(0) path; the clear is on the consumed slot before the read position advances, not on the producer's next slot.


sp00nznet:

PR #84, opened 2026-07-25
cellGame: real parameter IDs, error values and ABI buffer bounds

canersaka:
cellGame's integer parameter table was not the firmware ABI. Several IDs were shifted or collapsed together, VERSION, PS3_SYSTEM_VER and APP_VER did not have their distinct values, localized title IDs were incomplete, and the path and directory-name limits were host-shaped rather than the fixed guest sizes. A title asking for a valid parameter could therefore receive the value for a different key, while callers allocating the documented buffers could be overrun by the larger local limits. The module error base was also 0x8002B600 even though cellGame uses the 0x8002CBxx range.

This replaces the table with the firmware IDs through 106, restores the localized-title ranges and separate version fields, uses the 128-byte path and 32-byte directory-name bounds, corrects the cellGame and game-data error bases, and reads the three version strings from their corresponding SFO keys. The change is limited to constants, bounds and the existing parameter switch; it does not add title-specific behavior.

Verified: ps3recomp_runtime builds clean. A standalone compile-time ABI check passes for the corrected constants and buffer sizes. The numeric IDs, 0x8002CBxx error range, 128-byte path and 32-byte directory name were checked against RPCS3's cellGame.h as an ABI oracle.


sp00nznet:

PR #85, opened 2026-07-25
cellFs: Readdir writes the 258 byte dirent ABI and real CellOS errors

canersaka:
cellFsReaddir was writing CellFsDirectoryEntry, the separate stat-carrying shape used by cellFsGetDirectoryEntries, into an API whose output is the compact CellFsDirent `{ d_type, d_namlen, d_name[256] }`. It also returned nread=1 instead of the number of bytes written. A guest using the real 258-byte allocation had the wrong fields at every offset and could have the larger host-side structure written past its buffer.

This gives Readdir its real packed 258-byte shape on both Win32 and POSIX, fills d_type and d_namlen, and reports 0x102 in nread. The prerequisite commit also corrects the generic CellOS error table from 0x8001001A through 0x8001003E; the old table had EBADF at 0x80010025 and then aliased EFSSPECIFIC onto the same value, so cellFs returned numbers real guest code would not recognize. cellFsGetDirectoryEntries and its stat-carrying structure are deliberately left alone.

Verified: ps3recomp_runtime builds clean. CellFsDirent has a compile-time size assertion of 258 bytes, both host paths write that exact shape, EOF still returns nread=0, and the generic error values were checked against RPCS3's ErrorCodes.h as an ABI oracle.


sp00nznet:

PR #86, opened 2026-07-25
ppu: guest callbacks receive r3 through r10

canersaka:
The guest-call bridge only accepted seven integer arguments, r3 through r9, even though the 64-bit PowerPC ABI has eight volatile argument registers through r10. Every wrapper and callback adapter inherited that truncation. A callback with an eighth register argument silently received zero or stale state, and the trophy callback path made the loss visible by explicitly discarding its fifth callback argument before entering the guest.

This extends the shared dispatcher and its wrappers to carry r3 through r10, then updates the existing call sites mechanically. Shorter callbacks still pass zero in unused positions. The trophy adapter now forwards the argument it previously discarded; there is no per-title special case.

Verified: ps3recomp_runtime builds clean with all twelve call sites updated, and the PPU dispatcher translation unit compiles independently with the new signature. The diff is signature propagation plus the one trophy forwarding correction; no callback selection or return-value behavior changes.


sp00nznet:

PR #87, opened 2026-07-25
spu: sync, dsync and syncc emit real host memory fences

canersaka:
sync, dsync and syncc were lifted as no-ops, and the disassembler folded sync's C bit into the same plain mnemonic. Guest SPU code uses these instructions to order local-store, channel and DMA-visible state; dropping them gives the host compiler and CPU permission to move accesses across a barrier that the guest relied on, so the failure only appears under contention.

This decodes the C bit instead of discarding it and routes all three forms through the cross-compiler fence helper. sync uses the full fence, dsync the data-ordering fence and syncc the channel-ordering fence. No ordinary load, store or channel path changes.

Verified: the exact instruction words 0x00400000, 0x00500000 and 0x00600000 decode to their distinct forms and lift to real fences, and ps3recomp_runtime builds clean from the current ps3recomp staging base.


sp00nznet:

PR #88, opened 2026-07-25
lv2: keep one authoritative semaphore count

canersaka:
The Win32 semaphore path kept the guest count twice: once in `sys_semaphore_info::value` and again in the kernel semaphore handle. wait and trywait consumed the kernel count and then decremented value separately, while post updated value before releasing the handle. Timeout and wake races could make those counts disagree, producing lost tokens, negative bookkeeping or a waiter sleeping while value was already positive. The overflow path also returned CELL_EINVAL instead of the lv2 CELL_EBUSY contract.

This makes value under value_lock the only count on both host paths. The Win32 handle becomes a wake channel only, parked waiters recheck value under the lock, and post wakes at most `min(count, waiters)`. The second commit wires the existing sync_stress target into the test build with the small standalone shims it needs, so this behavior stays executable outside a generated title.

Verified: ps3recomp_runtime builds clean and CTest passes the sync_stress target, 1 of 1 tests. On the unmodified base the same semaphore-counting section reports ten failures; on this branch its post, trywait, timed-wait and overflow checks all pass.


sp00nznet:

PR #82, opened 2026-07-24
[DRAFT] integrate/faithful-adopt-caner: SPU lifter faithful adoption + LBP SPU/SPURS + RSX textures

sp00nznet:
**Attribution:** this branch is a from-scratch re-derivation of the SPU decoder/lifter + runtime originally authored by [@sp00nznet](https://github.com/sp00nznet) and [@canersaka](https://github.com/canersaka); credit for the underlying design belongs to them (their original commits stand on their own). sagemono authored the re-adoption. See CONTRIBUTORS.md.

**DRAFT -- compile-verified, NOT yet re-lift/run verified.** Do not merge before the validation below.

Folds `sagemono/spu/faithful-adopt-caner` (150 commits) on top of #81 (integrate/fold-2026-07-24). Brings a faithful re-adoption of caner's decode/lift work + LBP SPU/SPURS bring-up + RSX texture support (G8B8/DXT, deswizzle). -- [@sagemono](https://github.com/sagemono), adopting [@canersaka](https://github.com/canersaka)'s decode work.

## Why this was hard (and the resolution principle)

The two lineages independently evolved the SAME subsystems (PPU/SPU lifter, FP/VMX handlers, SPU runtime, HLE) with DIFFERENT helper vocabularies -- e.g. lane-safe BE VMX floats via `vrf`/`vstf` (ours) vs `ppu_vldf4`/`ppu_vstf4` (theirs); `_mem_base` rA=0 vs raw `ctx->gpr[base]`. Cherry-picking hunks across them would reference helpers defined on only one side. So this is a "which lineage is canonical" decision, not a hunk merge. 72 conflict hunks across 23 files.

**Canonical = sagemono's faithful-adopt-caner** (the deliberate, curated adoption). Resolution:
- **Lifter tools** (`ppu_lifter.py` + spu_lifter/spu_disasm/gen_hle_nids): took theirs, then RE-APPLIED our jump-table-base fix (their `discover_jump_tables` had the identical "first TOC-loaded operand wins" bug that dropped flow's app-loop dispatcher).
- **SPU runtime + HLE** (spu_context/dma/channels/workload, cellSpurs, cellSaveData, ppu_sysprx, ppu_loader): took theirs (canonical engine), but PRESERVED flow-critical ydkj functional bits: `ydkj_host_bt` re-added (ppu_fs.cpp calls it), `ydkj_release_pending_threads` unioned (sys_event.c calls it), `park_on_empty_inmbox`, `TUNERFIX`.
[...]

PR #90, opened 2026-08-10
Repair the faithful-adopt-caner merge: 3 dropped lifter fixes + 6 dangling runtime refs

sp00nznet:
Fixes for regressions introduced by the `faithful-adopt-caner` merge itself. Targets #82's branch so that PR isn't rewritten — merge or cherry-pick, whichever suits.

## The merge resolution for `tools/ppu_lifter.py` was not a superset

```
git diff sagemono/spu/faithful-adopt-caner HEAD -- tools/ppu_lifter.py   # 28 lines
```

It took sagemono's file wholesale and discarded ~835 lines of the ydkj side. Three of the losses are silent miscompiles:

| dropped | evidence in a fresh flOw lift |
|---|---|
| `ctx->lr = <ret>` on direct `bl` | every `lr=` field in the runtime logs printed `0x00000000`; `ctx->lr` writes 140k → 29k |
| TOCFIX `ld r2,N(r1)` → literal module TOC | **14,146** sites reloading r2 from uninitialised stack → garbage TOC → OPD loads read code-as-data |
| `_mem_base()` rA=0 D-form literal-zero base | loads/stores basing on `ctx->gpr[0]` (the `_xea()` X-form twin survived) |

After relifting: 14146 TOCFIX / 0 stack-TOC reads / 120101 `ctx->lr` writes / 0 `gpr[0]` D-form bases, and the `lr=` diagnostics match the known-good ps3-draw lift exactly.

**Still dropped, not restored here** (flOw's EBOOT doesn't use them, so I can't verify them): VRSAVE `mfspr`/`mtspr`, `stvlx/stvrx(l)`, `vrfin/vrfiz/vrfip/vrfim`, `vexptefp`, `vlogefp`. Worth a follow-up.

## The runtime didn't link for any title

Undefined symbols in a `.lib` are only diagnosed when a title links it, so the archive built clean and this went unnoticed.

The SPU interpreter was **not** "consolidated into `spu_channels.c`" as 6dd8524 says — it was deleted, while both call sites in `lv2_register.c` stayed. Restored, along with `spu_run_interp_job`.

The *function registry* genuinely was consolidated, so restoring `spu_fn_registry.c` wholesale duplicates `spu_register_function`/`spu_lookup`/`spu_begin_image`; only `spu_lifted_lookup` was missing and it's re-added as the 3-line wrapper it always was.

`lbp_hle_complete_pending`, `g_taskset_policy_bytes/size` and `g_vm_page_bitmap` are title symbols referenced from shared runtime code. Each gets a default definition in its own **one-symbol TU**, per the existing `spu_tsp_weak.c` pattern — a port supplying the real one never pulls the archive member. They must stay separate files; a port can supply one without the others. `g_vm_page_bitmap` storage moves out of `lbp/main.cpp`, which is precisely why every non-LBP port failed.

> Gotcha: the runtime CMakeLists globs `runtime/*.c` without `CONFIGURE_DEPENDS` — a new TU needs `cmake -S . -B build` re-run or it's silently never compiled.

## Two more dropped/missing behaviours

- **`cellSaveData`**: ps3-draw returned `CELL_OK` on `CBRESULT_ERR_NODATA` with the comment *"an ERROR return leaves the title parked in MODE_AUTO_LOAD"*. The merge restored the spec-correct `CELL_SAVEDATA_ERROR_NODATA` and flOw parked exactly as predicted — 0 flips, never reaching `m_InitEntityHierarchy`. Applied to **both** entry points: flOw calls the old non-`_2` `cellSaveDataAutoLoad`, which 9db188d added as a separate function, so patching only `AutoLoad2` does nothing. Per the SDK the error return is correct; that flOw can't handle it is an unexplained bug in its own state machine, and the code now says so.
- **`sys_ppu_thread`**: guest threads never got `SetThreadStackGuarantee`, so on a stack overflow the handler faults while reporting and the process dies with a bare AV *inside the handler*, losing the backtrace. `main()` already did this for the main thread.

## Status

flOw builds and boots substantially further against this branch (209 → 306 log lines: savedata clears, `thr_auto_load()` completes), then hits a stack overflow before the app loop. `FLOW_NO_SPU_WL=1` reproduces it, so it's the PPU lift. Lead for the follow-up: this lifter emits **7,915** `g_trampoline_fn` sites vs ps3-draw's **19,171**, and 16 mid-function tail-entry wrappers vs 1,141, at near-identical function counts (38,287 vs 38,421) — so it's the branch *routing*, not the function list. The trampoline exists specifically to stop backward branches re-entering a prologue and growing the stack.


sp00nznet:
Merged into master via the integration rounds — the branch is an ancestor of master, so this is closed as landed rather than dropped.

PR #92, opened 2026-08-15
Fix post-#91 master: clang-cl build, two missed guest-call sites, dropped SPU park

sp00nznet:
Four fixes needed to build and run a real title (the Rubber Ducky port) against
`master` after #91. Two are build breaks, two are behavioural regressions the
fold itself introduced.

### 1. clang-cl build breaks (pre-existing)

- **`getenv` without `<stdlib.h>`** in `libs/spurs/cellSpurs.c`, `libs/video/rsx_commands.c`,
  `runtime/syscalls/sys_timer.c`. clang-cl errors on the implicit declaration; MSVC
  only warns (`C4013`) and then silently truncates the returned `char*` to `int` —
  a live 64-bit bug on the MSVC path, not clang pedantry.
- **Stale local `extern`** for `spu_run_lifted_job_abi` in `runtime/spu/spu_workload.c`,
  conflicting with the `static inline` in `spu_lifted_job.h` that the same file
  already includes.

Both were fixed locally in the Rubber Ducky checkout months ago (its README lists
them under "Research-branch fixes carried in this port") and never pushed upstream.

### 2. Two guest-call sites #86's r3–r10 widening missed

`#86` widened `ppu_guest_call` / the `g_ps3_guest_caller` hook to nine arguments and
updated every call site on its base. Two sites added later by the integration chain
were not on that base:

- `runtime/ppu/ppu_hle.cpp:407` — `ppu_guest_call(iopd, desc, 1, 8, 0x398)`
- `lbp/main.cpp:174` — `harness_guest_caller` plus a local 5-param `ps3_guest_caller_fn`
  typedef and `extern`. This one is worse than a compile error: it installed a
  five-parameter function as the hook, so every nine-argument call through
  `g_ps3_guest_caller` was an ABI mismatch.

These did not show up in #91's verification because `ppu_hle.cpp` and `lbp/main.cpp`
are per-title files — not part of `ps3recomp_runtime` or the `sync_stress` target.

### 3. `park_on_empty_inmbox` dropped by the runtime/spu consolidation

`eeff394` added `spu_context.park_on_empty_inmbox`, letting a persistent-worker raw
SPU run synchronously (full init + ready-mailbox handshake, then park) instead of on
an async host thread racing the PPU. `6dd8524` ("fold: align runtime/spu with
sagemono's consolidated structure") dropped both the field and the `rchcnt` behaviour,
silently un-fixing the determinism it was written for. Restored into the consolidated
`spu_channels.c` / `spu_context.h`.

The same commit's `RD_FORCE_ADDR` read-override diagnostic in `ppu_loader.cpp` is
still missing; left out here since it is debug-only.

## Verification

- `ps3recomp_runtime.lib` links clean under clang-cl 21.1.8 (Ninja, Release).
- The Rubber Ducky port re-lifts, builds (22.4 MB exe) and runs against this branch,
  reaching the same point as its last known-good build.

### DIFF
--- a/docs/broad-testing-findings.md
+++ b/docs/broad-testing-findings.md
@@ -8,7 +8,7 @@ git-ignored; this file is the human summary.)
 
 ## 1. PSN library catalog (filename pass, no extraction)
 
-- **1,345 archives** under `<PSN_ROOT>` (PSN_1: 672, PSN_2: 673), **~1.13 TB** total.
+- **1,345 archives** in the local PSN library (set `PS3RECOMP_PSN_ROOT`; two shards of 672 / 673), **~1.13 TB** total.
 - **1,339 / 1,345** carry a parseable title-id in the filename (`… [NPxx-NNNNN]`).
 - **Region:** EU (SCEE) 661 · US (SCEA) 635 · JP (SCEJ) 28 · Asia 6 · unknown 15.
 - **Content class** (PSN prefix 4th letter, coarse):
--- a/lbp/main.cpp
+++ b/lbp/main.cpp
@@ -165,14 +165,18 @@ extern "C" void lv2_init_syscalls(void);   /* runtime/syscalls/lv2_register.c */
  * calling cellGcmTickVBlank()/TickFlip(), which invoke the registered handlers.
  * Without this the game inits, registers its handlers, and then waits forever
  * for a vblank that never comes. */
-typedef void (*ps3_guest_caller_fn)(uint32_t, uint64_t, uint64_t, uint64_t, uint64_t);
+typedef void (*ps3_guest_caller_fn)(uint32_t, uint64_t, uint64_t, uint64_t, uint64_t,
+                                    uint64_t, uint64_t, uint64_t, uint64_t);
 extern "C" ps3_guest_caller_fn g_ps3_guest_caller;        /* libs/system/cellSysutil.c */
-extern "C" uint64_t ppu_guest_call(uint32_t, uint64_t, uint64_t, uint64_t, uint64_t);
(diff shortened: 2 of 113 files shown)


sp00nznet:
### Result on Rubber Ducky: the Cg shader wall is gone

The 4th dropped fix (`dee21c4`, restoring 27982e0's sldi+add+lwz jump-table idiom) clears the wall that port had been stuck on since v0.18.

`_jsCountFloatsInCgType`'s 110-case switch sits inside the Cg shader loader. Without the idiom, `discover_jump_tables` bailed (`JT_DEBUG` showed `lwzx=None`), the `bctr` lifted to a bare `ps3_indirect_call`, and at runtime it logged `unresolved indirect call -> 0x00143DBC` and returned without dispatching — so `_jsCreatePushBuffer` got a garbage program layout and Cg rejected the handle.

**Lift:** 204 → 252 dispatchers, 2042 → 2458 case targets (detection unchanged at 8530 functions, so this is purely codegen).

**Run, same binary and assets:**

| | before | after |
|---|---|---|
| unresolved indirect calls | 4 | **0** |
| `.vpo` shaders loaded | 1 | **8** (every shader) |
| meshes read | 3 | **7** |
| `Started solver thread` | no | **yes** |
| abort | `cg_particlefluid.cpp:101 g_paramProjection` | **none** |

It now runs to the 150s bound instead of aborting. New park point is `TclMultiSim::initialize+0x78` → `sys_event_queue_receive(q=9, timeout=0)`, waiting on an SPU thread group that reports `no fallback for any of 1 thread(s); instantly completed` — an unregistered SPU workload, a different and much later wall.

PR #95, opened 2026-08-23
build: make the runtime library compile on arm64 macOS (+ macOS CI)

### LINKED ISSUE #94: macOS is documented as supported but does not compile (77 errors on arm64); fix + CI ready
## Summary

`docs/BUILDING.md` lists Apple Clang 14+ under **"Full support"** for macOS, and `docs/FAQ.md` says macOS "should work but may need minor fixes". In practice the tree does not compile on macOS at all.

On an Apple Silicon host (M3 Max, macOS 15, Apple Clang 21, CMake 4.4, Ninja), CMake **configures cleanly** and then the build produces **77 errors across 14 files**.

I have a working fix and would like to contribute it — opening this first to check the approach is one you'd want before sending the PR.

## Reproduction

```
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug \
      -DCMAKE_C_FLAGS=-ferror-limit=0 -DCMAKE_CXX_FLAGS=-ferror-limit=0
ninja -C build -k 0
```

`-ferror-limit=0` matters: with clang's default limit the tail is truncated and you see ~63. Also note that fixing the one *fatal* error (`intrin.h`) causes `spu_channels.c` to go from 0 visible errors to 15, because a fatal error aborts the TU and masks everything behind it.

## Error inventory

| File | Errors |
|---|---|
| `libs/video/cellGcmSys.c` | 29 |
| `runtime/spu/spu_channels.c` | 15 (masked behind a fatal until `intrin.h` is fixed) |
| `libs/spurs/cellFiber.c` | 8 |
| `libs/system/cellSaveData.c` | 6 |
| `libs/spurs/cellSpurs.c` | 6 |
| `runtime/ppu/ppu_memory.h` (via `libs/spurs/`) | 4 |
| `libs/filesystem/cellFs.c` | 3 |
| `libs/video/rsx_commands.c` | 10 |
| `runtime/syscalls/sys_fs.c` | 2 |
| `libs/sync/cellSync2.c` | 2 |
| `runtime/syscalls/{sys_timer,sys_mutex}.c` | 2 |
| `runtime/spu/spu_workload.c` | 1 |
| `libs/{input/cellPad,audio/cellAudio}.c` | 2 (`SDL2/SDL.h` not found) |

By kind: 35 implicit function declarations, 14 undeclared identifiers, 9 missing struct members, 6 unknown type names, 4 static-vs-non-static conflicts, 3 missing headers.

## Root causes

1. **Missing includes** — `getenv` (×10) and `errno` (×2) used without `<stdlib.h>`/`<errno.h>`.
2. **Darwin stat fields** — the `#else` branch assumes glibc's `st_atim`/`st_mtim`/`st_ctim`; Darwin spells these `st_*timespec`.
3. **ucontext** — Darwin hides `makecontext`/`swapcontext`/`setcontext` unless `_XOPEN_SOURCE` is defined before the header.
4. **MSVC intrinsic unguarded** — `spu_channels.c:228` includes `<intrin.h>` and uses `_InterlockedExchange` with no `_MSC_VER` guard.
5. **Win32 sync/timing in the shared path** — `cellGcmSys.c`, `spu_channels.c` and `cellSpurs.c` use `SRWLOCK`, `CONDITION_VARIABLE`, `QueryPerformanceCounter`, `CreateEventA`, `CreateThread`, `Sleep` directly. 18 other files already hand-roll a `_WIN32`/pthread split; these just missed it.
6. **Absent on macOS** — `pthread_mutex_timedlock` and `sem_timedwait` are not provided by Darwin.
7. **CMake** — never calls `find_package(SDL2)` off Windows, although `cellPad`/`cellAudio` select their SDL2 backends unconditionally there (`PAD_BACKEND_SDL2` / `AUDIO_BACKEND_SDL2`).

## Three defects that are not macOS-specific

A second toolchain surfaced these; they are latent on every platform:

- **`libs/spurs/spurs_taskset.h:25-28`** re-declares `vm_read32`/`vm_read64`/`vm_write32`/`vm_write64` **without** `static`, contradicting the `static inline` definitions in `runtime/ppu/ppu_memory.h`. The header comment says they are "defined in runtime/ppu/ppu_loader.cpp", so there appear to be two competing definitions. MSVC accepts this; Clang rejects it. **Which one is intended?**
- **`runtime/spu/spu_workload.c:421`** redeclares `spu_run_lifted_job_abi` as `extern` with a raw function-pointer type, conflicting with the `static inline` definition in `spu_lifted_job.h`, which is already included at line 9.
- **`libs/video/rsx_commands.c:548`** assigns an implicitly-declared `getenv()` to a `const char*`. With no declaration the return type is `int`, so **the pointer is truncated to 32 bits**. This is a real bug on any 64-bit target; MSVC only avoids it by declaring `getenv` transitively.

## Proposed fix

- Targeted portability fixes for causes 1–4 and 7.
- A small `runtime/platform/win32_compat.h` supplying the Win32 sync/timing names on POSIX, so the existing call sites stay unchanged, plus `darwin_compat.h` for the two missing POSIX functions.
  - One subtlety worth calling out: `spu_context` stores `ch_wait_lock`/`ch_wait_cv` in bare `void*` slots because Win32 `SRWLOCK`/`CONDITION_VARIABLE` are pointer-sized. `pthread_mutex_t` is 64 bytes on Darwin and `pthread_cond_t` 48, so a typedef-based shim would overflow into adjacent struct members. The shim keeps `void*` storage and lazily heap-allocates the pthread object into the slot with an atomic CAS, so `NULL` still means "uninitialised" and existing `memset` zero-init stays valid.
- A macOS arm64 CI job (Debug + Release). There is currently no CI in the repo, which is plausibly why the POSIX paths drifted.

Result: clean Debug and Release builds, `libps3recomp_runtime.a` is arm64, and `runtime/spu/tests/test_spu_helpers.c` reports **63 passed, 0 failed**.

## Two things I would leave out of that PR

- **`tests/sync_stress`** is written directly against Win32 threading (`WaitForMultipleObjects`, joinable `HANDLE`s, `Interlocked*`, `CRITICAL_SECTION`). Porting it needs a real handle table; I would gate it to `WIN32` and leave it as follow-up. Happy to do it separately if you'd prefer it ported now.
- **`libs/video/tests/test_fp_decompiler.c`** calls `rsx_fp_decompile` with 4 arguments, but it takes 5 since `exports32` was added — so it does not compile. Fixing the call sites is trivial, but its 9 assertions are then stale against the current emitter and **all 9 fail** with `exports32` at either 0 or 1 (the expected `r[0].xyzw =` no longer matches what the emitter produces). Since these tests are not wired into CMake, they appear never to have run. I did not want to guess at the intended expectations — is the emitter output correct and the test stale, or the reverse?

Happy to split any of this into smaller PRs if that's easier to review.

slushiimusic:
Fixes #94.

`docs/BUILDING.md` lists Apple Clang 14+ under **"Full support"** for macOS, but the tree did not compile there at all — a clean configure followed by a build produced **77 errors across 14 files**. This gets it to a green build on Apple Silicon in both Debug and Release.

Verified on an M3 Max (macOS 15, Apple Clang 21, CMake 4.4, Ninja, 16 KB pages).

## Portability fixes

| Cause | Fix |
|---|---|
| `getenv`/`errno` used without `<stdlib.h>`/`<errno.h>` | includes added (3 files) |
| `#else` branch assumed glibc's `st_atim`/`st_mtim`/`st_ctim` | `__APPLE__` branch using `st_*timespec` (3 sites) |
| Darwin hides the ucontext routines | `_XOPEN_SOURCE` before the header in `cellFiber.c` |
| `spu_channels.c:228` used MSVC `_InterlockedExchange` unguarded | C11 `stdatomic` off MSVC, matching the ordering used elsewhere in-tree |
| CMake never looked for SDL2 off Windows | `find_package(SDL2)` + `Threads`, with an actionable message if missing |

## New: `runtime/platform/`

18 files already hand-roll a `_WIN32`/pthread split. `cellGcmSys.c`, `spu_channels.c` and `cellSpurs.c` did not, and used `SRWLOCK`, `CONDITION_VARIABLE`, `QueryPerformanceCounter`, `CreateEventA`, `CreateThread` and `Sleep` directly. `win32_compat.h` supplies those names on POSIX so the call sites are unchanged; on Windows it is a passthrough to `<windows.h>`.

`darwin_compat.h` supplies `pthread_mutex_timedlock` and `sem_timedwait`, which macOS does not ship, by polling the non-blocking variant against the caller's deadline.

> **Storage-size subtlety.** `spu_context` stores `ch_wait_lock`/`ch_wait_cv` in bare `void*` slots because Win32 `SRWLOCK` and `CONDITION_VARIABLE` are pointer-sized and zero-initialise validly. `pthread_mutex_t` is 64 B on Darwin and `pthread_cond_t` 48 B, so a typedef-based shim would write straight through the slot into adjacent struct members. The shim keeps `void*` storage and lazily heap-allocates the pthread object into the slot with an atomic CAS — `NULL` still means "uninitialised", so the existing `memset` zero-init stays correct.

## Three defects surfaced by a second toolchain

These are latent on every platform, not macOS-specific:

- **`spurs_taskset.h`** re-declared `vm_read32`/`vm_read64`/`vm_write32`/`vm_write64` without `static`, contradicting the `static inline` definitions in `ppu_memory.h`. MSVC accepted it; Clang rejects it. Now includes the canonical header — but see the question in #94 about which definition is intended.
- **`spu_workload.c:421`** redeclared `spu_run_lifted_job_abi` as `extern` with a raw function-pointer type, conflicting with the `static inline` definition in `spu_lifted_job.h` (already included at line 9). Removed.
- **`rsx_commands.c:548`** assigned an implicitly-declared `getenv()` to a `const char*`. With no declaration the return type is `int`, so **the pointer was truncated to 32 bits** — a real bug on any 64-bit target. Fixed by the `<stdlib.h>` include.

## CI

Adds a macOS arm64 job (Debug + Release). There was no CI in the repo, which is plausibly why the POSIX paths drifted; this is what keeps them from drifting again. It also runs `test_spu_helpers` as a smoke test.

## Verification

- Clean-room Debug **and** Release builds: 0 errors
- `libps3recomp_runtime.a` — `lipo -info` reports `arm64`, 121 objects
- `runtime/spu/tests/test_spu_helpers.c`: **63 passed, 0 failed**

## Deliberately not included

- **`tests/sync_stress`** is Win32-only (`WaitForMultipleObjects`, joinable handles, `Interlocked*`, `CRITICAL_SECTION`). Gated to `WIN32` with the reason in-tree; porting it needs a real handle table. Happy to do that separately.
- **`test_fp_decompiler.c`** called `rsx_fp_decompile` with 4 args after it gained `exports32`. The 6 stale call sites are fixed here so it compiles, but its 9 assertions are stale against the current emitter and all fail with `exports32` at either 0 or 1. These tests are not wired into CMake and appear never to have run. I did not want to guess at the intended expectations — see #94.

Happy to split this into smaller PRs if that's easier to review.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -1,0 +1,45 @@
+# Builds the runtime library on Apple Silicon.
+#
+# ps3recomp had no CI of any kind, and docs/BUILDING.md advertised macOS as
+# "Full support" while the tree did not compile there at all. This job exists so
+# that stays fixed: it is the only thing preventing the next Windows-only commit
+# from silently re-breaking the POSIX paths.
+name: macOS
+
+on:
+  push:
+    branches: [ master, main ]
+  pull_request:
+  workflow_dispatch:
+
+jobs:
+  build:
+    name: macOS arm64 (${{ matrix.build_type }})
+    runs-on: macos-14          # Apple Silicon runner
+    strategy:
+      fail-fast: false
+      matrix:
+        build_type: [Debug, Release]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: brew install ninja sdl2
+
+      - name: Configure
+        run: cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=${{ matrix.build_type }}
+
+      - name: Build
+        run: cmake --build build
+
+      - name: Verify architecture
+        run: |
+          lipo -info build/libps3recomp_runtime.a
+          lipo -info build/libps3recomp_runtime.a | grep -q arm64
+
+      - name: Run self-contained SPU helper tests
+        run: |
+          clang -std=gnu17 -I include -I runtime/spu \
+                -o /tmp/test_spu_helpers runtime/spu/tests/test_spu_helpers.c
+          /tmp/test_spu_helpers
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,31 @@ if(WIN32)
         ole32       # COM for WASAPI audio
         bcrypt      # BCryptGenRandom for SSL RNG
     )
+else()
+    # cellPad and cellAudio select their SDL2 backends unconditionally off
+    # Windows (see PAD_BACKEND_SDL2 / AUDIO_BACKEND_SDL2), so SDL2 is required
+    # here -- docs/BUILDING.md already lists it as the Linux/macOS dependency.
+    find_package(SDL2 QUIET)
+    if(SDL2_FOUND)
+        if(TARGET SDL2::SDL2)
+            target_link_libraries(ps3recomp_runtime PRIVATE SDL2::SDL2)
+        else()
+            target_include_directories(ps3recomp_runtime PRIVATE ${SDL2_INCLUDE_DIRS})
+            target_link_libraries(ps3recomp_runtime PRIVATE ${SDL2_LIBRARIES})
+        endif()
+        message(STATUS "  SDL2:         found (input + audio backends enabled)")
+    else()
+        message(FATAL_ERROR
+            "SDL2 not found. It provides the cellPad and cellAudio backends on "
+            "non-Windows hosts.\n"
+            "  macOS:  brew install sdl2\n"
+            "  Debian: apt install libsdl2-dev\n"
+            "If SDL2 is installed in a non-standard prefix, pass "
+            "-DCMAKE_PREFIX_PATH=<prefix>.")
+    endif()
+
+    find_package(Threads REQUIRED)
+    target_link_libraries(ps3recomp_runtime PRIVATE Threads::Threads)
 endif()
 
 # ---------------------------------------------------------------------------
@@ -117,7 +142,16 @@ endif()
 # ---------------------------------------------------------------------------
 if(PS3RECOMP_BUILD_TESTS)
     include(CTest)
-    add_subdirectory(tests/sync_stress)
+    # sync_stress is written directly against the Win32 threading API
+    # (WaitForMultipleObjects, joinable HANDLEs, Interlocked*, CRITICAL_SECTION).
+    # runtime/platform/win32_compat.h covers the subset the runtime itself needs,
+    # but not joinable handles or multi-object waits -- porting this test needs a
+    # real handle table and is left as follow-up work.
+    if(WIN32)
+        add_subdirectory(tests/sync_stress)
+    else()
+        message(STATUS "  sync_stress:  skipped (Win32-only threading API)")
+    endif()
 endif()
 
 # ---------------------------------------------------------------------------
--- a/libs/filesystem/cellFs.c
+++ b/libs/filesystem/cellFs.c
@@ -216,6 +216,11 @@ static void fill_cellfs_stat(CellFsStat* sb, const HOST_STAT_T* hst)
     sb->st_atime = (s64)hst->st_atime;
     sb->st_mtime = (s64)hst->st_mtime;
     sb->st_ctime = (s64)hst->st_ctime;
+#elif defined(__APPLE__)
+/* Darwin names the sub-second stat fields st_*timespec, not glibc's st_*tim. */
+    sb->st_atime = (s64)hst->st_atimespec.tv_sec;
+    sb->st_mtime = (s64)hst->st_mtimespec.tv_sec;
+    sb->st_ctime = (s64)hst->st_ctimespec.tv_sec;
 #else
     sb->st_atime = (s64)hst->st_atim.tv_sec;
     sb->st_mtime = (s64)hst->st_mtim.tv_sec;
--- a/libs/spurs/cellFiber.c
+++ b/libs/spurs/cellFiber.c
@@ -14,6 +14,11 @@
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #else
+/* Darwin marks the ucontext routines deprecated and hides them unless
+ * _XOPEN_SOURCE is defined before the header is pulled in. */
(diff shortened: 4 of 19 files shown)


sp00nznet:
Merged — thanks, and sorry for the silent merge.

You were right that `docs/BUILDING.md` claimed macOS support the tree couldn't deliver. Two things needed follow-ups on master after the merge, both caught by your CI within 17 seconds of it landing:

- `libs/codec/cellSail.c` called `getenv` without `<stdlib.h>` — Clang errors on the implicit declaration MSVC accepted.
- `runtime/syscalls/lv2_register.c` used `LARGE_INTEGER` / `QueryPerformanceCounter` unguarded, for the `SPU_SPEED` diagnostic that measures interpreter throughput. Rather than `#ifdef` the diagnostic away and lose it on macOS, I shimmed the three names onto `CLOCK_MONOTONIC` in your `darwin_compat.h`.

I swept for both patterns rather than fix-and-wait on CI: `cellSail.c` was the only unguarded `getenv`, and of the seven `QueryPerformance*` users five were already inside `#ifdef _WIN32` with the whole D3D12 backend wrapped from its line 17. Third run green.

Worth stating plainly for anyone reading this later: **this makes the runtime *library* build on macOS — it does not make anything render.** Both backends (`rsx_d3d12_backend.c`, `rsx_null_backend.c`) are still `#ifdef _WIN32`, there's no POSIX host layer, and the CI job builds `libps3recomp_runtime.a` and runs one SPU helper test without ever linking an executable. That's a genuine step — it closes the docs-vs-reality gap you filed #94 about, and it's this repo's first CI — but a macOS user still can't run anything yet.

That's what #96 is for, and it's next on my list.

PR #96, opened 2026-08-23
video: RSX -> Metal backend, POSIX host, and the draw path

slushiimusic:
> **Stacked on #95.** That PR makes the tree compile on arm64 macOS; this one needs it. Until #95 merges, GitHub will show both commits here — review only the second (`video: add an RSX -> Metal backend and a POSIX host harness`).

Gives macOS its first working render path.

Today the only backend that draws anything is D3D12, which is Windows-only. `rsx_null_backend.c`'s non-Windows branch is a stub that prints "not available" and returns `-1`, carrying a standing `TODO: SDL2 or X11 backend`. This is that TODO.

## `libs/video/rsx_metal_backend.{h,m}`

Implements the two `rsx_backend` callbacks a guest clear/flip loop actually needs — `clear` captures the colour from `NV4097_CLEAR_SURFACE`, `set_render_target` tracks the guest surface size. The other 15 stay `NULL`: every dispatch site in `rsx_commands.c` is guarded (`if (s_backend && s_backend->x)`), so a partial vtable is the intended way to bring a backend up incrementally, the same shape `libs/video/tests/test_rsx_fifo.c` already relies on.

Windowed through a `CAMetalLayer` on an `NSWindow`, or headless to an offscreen `MTLTexture` under `PS3RECOMP_METAL_HEADLESS=1`. Headless exists for two reasons: it makes the backend verifiable in CI on a machine with no display, and it lets the presented pixel be read back off the GPU and asserted.

**Metal rather than SDL_Renderer, deliberately.** SDL2's renderer is a 2D sprite API — no custom vertex program, no depth/stencil, no MRT, no render-to-texture. It can produce a first pixel but cannot grow into the real pipeline, and the work would be thrown away.

## `runtime/host/{host_posix.c,host_stubs.c}`

A host that drives `cellGcm` → RSX → backend **with no lifted game**: it hand-builds a real big-endian NV4097 command buffer, feeds it to the FIFO, and runs the flip loop. No `ppu_recomp.h`, no function table, no recompiler run. The HLE library needs exactly five out-of-line symbols from the PPU loader, supplied inert in `host_stubs.c` (~14 lines).

This is the part that unblocks people. Non-Windows graphics work has been hard to *start* because it looked like it required a game binary and a full lift first. It doesn't. The host exits non-zero if the guest command stream doesn't survive the trip, so it also works as a regression check.

## Verification

Clean Debug **and** Release builds on an M3 Max, 0 errors, 0 new warnings:

```
[RSX metal] headless on Apple M3 Max (1280x720)
[RSX] CLEAR_SURFACE mask=0xF0 color=0xFF101830
[cellGcmSys] Init(cmdSize=0x10000, ioSize=0x200000, ioAddr=0x00100000)
[host] gcm context data @ 0x00400010
[cellGcmSys] SetDisplayBuffer(id=0, offset=0x0, pitch=5120, 1280x720)
[host] clear colour through the FIFO: 0xFF101830 (expected 0xFF101830) OK
[host] presented 1 frame(s)
[host] presented pixel: 0xFF101830 (clear 0xFF101830) OK
```

That last line is a **GPU readback of what the Metal render pass actually produced** — an end-to-end assertion, not a compile check. The windowed path was verified separately against a live `CAMetalLayer`. The headless run is now a CI step in the macOS workflow.

## On the shader path (follow-up, not in this PR)

Worth recording, because it changes the cost estimate: **no MSL emitter is needed.** The existing `rsx_fp_decompiler.c` / `rsx_vp_decompiler.c` already emit HLSL, and `glslang` → SPIR-V → `spirv-cross` → `-newLibraryWithSource:` compiles that into a live `MTLRenderPipelineState` at runtime, with Command Line Tools only — no full Xcode, no offline `metallib`, no DXC.

Verified on this machine against the decompilers' exact token surface: `saturate`→`fast::clamp`, `lerp`→`mix`, `frac`→`fract`, `rsqrt`, `Texture2D`→`texture2d<float>`, `SamplerState`→`sampler`, `.Sample()`→`.sample()` all translate automatically, and the resulting pipeline renders and reads back correctly.

Two constraints found while proving it, for whoever wires the draw path:
- Both stages must go through the **same** `spirv-cross` run, or the `[[user(locn)]]` slots don't line up and PSO creation fails with `Fragment input(s) user(locn0),user(locn1) mismatching vertex shader output`.
- The vertex buffer needs a high bind index (30 works) because `buffer(0)` is taken by the constant buffer.

PSO cache and draw path are deliberately left out to keep this reviewable. Happy to split further if useful.

---

## Update: the draw path is in

A second commit takes this from clearing a drawable to **drawing guest geometry**.

- **Vertex fetch** ported from the D3D12 backend's `read_vp_vertex`: big-endian components, and the bit-31 context-DMA location on `NV4097_SET_VERTEX_DATA_ARRAY_OFFSET` selecting `cellGcmResolveOffset` (LOCAL) vs `cellGcmResolveLocated` (MAIN). Covers float32, half, u8-norm and s16-norm attribute types plus the instancing frequency divisor.
- **Primitive conversion** — Metal has no quads, quad strips, fans or polygons, exactly as D3D12 does not, so those expand to triangle lists during the fetch.
- **Pipeline-state cache** keyed on blend state, with RSX blend factors/equations mapped to `MTLBlendFactor`/`MTLBlendOperation`.
- **Deferred draw records** replayed inside one render pass at present, mirroring the D3D12 `D3D12DrawRecord`/`render_frame` split, because RSX draws arrive while the guest builds the frame and the flip comes later.

`host_posix --draw` uploads a clip-space triangle into the guest IO window, sets attributes 0 and 3 via `NV4097_SET_VERTEX_DATA_ARRAY_{OFFSET,FORMAT}`, and issues a real `SET_BEGIN_END`/`DRAW_ARRAYS` pair:

```
./ps3recomp_host                    -> pixel 0xFF101830 (clear)     OK
./ps3recomp_host --draw             -> pixel 0xFFFF0000 (triangle)  OK
./ps3recomp_host --draw --frames=60 -> pixel 0xFFFF0000 (triangle)  OK
```

All three are GPU readbacks, and all three run headless in CI.

**One finding worth flagging:** the harness originally rewound the command ring each frame and only the first frame submitted anything. `cellGcmSys` holds the RSX `get` pointer internally (`s_fifo_getoff`) and only reads `put` from guest memory, so a ring must be appended to and `put` advanced — which is how the hardware behaves. The runtime was right; the harness was wrong. Worth knowing for anyone else writing a host.

Still to come: textures, depth/stencil, offscreen render targets, and swapping the built-in shader for the translated-guest-shader path described above.


sp00nznet:
Merged. This is the piece that makes macOS mean something — #95 gave the tree a library that compiled and couldn't draw; this gives it a render path, and the headless GPU readback proves it rather than asserting an exit code.

Two follow-ups needed on master, both caused by a change I landed earlier today rather than by anything in this PR:

1. **Three undefined symbols at link.** `ppu_memory.h` expands every `vm_write*` into a range test against `g_ww_lo`/`g_ww_hi` plus a call to `ps3_ww_report_inline`. Those live in `runtime/ppu/ppu_loader.cpp`, which CMake excludes from the runtime library because it is compiled per-game — so a game link supplies them and `ps3recomp_host` did not. Stubbed inert in `host_stubs.c` (`lo == hi`, so the test never fires).

2. **Segfault before the first line of output.** Earlier today I moved the injected GCM block — label window, control register, IO↔EA offset tables — out of guest-allocatable memory to `VM_HLE_INJECT_BASE = 0x20000000`, because a title's own allocator was handing out the same addresses and clobbering it. Your host allocated a 64 MB VM, so `cellGcmSetupContext` wrote 512 MB past the end. It also hardcoded `CTRL_ADDR 0x03002000`, a stale copy of the old control address. Both now derive from `VM_HLE_INJECT_BASE`, so the block can move again in one edit — which is the same lesson `ppu_sysprx.cpp` taught me this morning when its private copy of the FIFO callback sentinel silently stopped matching.

Neither was your bug; the PR was written against the address the block used to live at.

CI is green on both configs and the assertions read as intended:

```
[host] clear colour through the FIFO: 0xFF101830 (expected 0xFF101830) OK
[host] presented pixel: 0xFF101830 (expected 0xFF101830) OK      # clear
[host] presented pixel: 0xFFFF0000 (expected 0xFFFF0000) OK      # --draw
```

On the shader note in your description — agreed, and it's worth doing in that order. Routing the existing `rsx_fp_decompiler.c` / `rsx_vp_decompiler.c` output through glslang → spirv-cross is a much smaller job than a second MSL emitter, and it keeps one shader path rather than two that drift.

sp00nznet:
Heads-up on credit, since you should have it: your commits are on `master` with authorship intact (`Julian Scanlan <julian@slushii.com>`) — the merge preserved them, nothing was squashed — but GitHub isn't attributing them to your profile, so you don't appear in the repo's Contributors sidebar.

The cause is just that `julian@slushii.com` isn't a verified email on your GitHub account, so there's nothing for GitHub to match the commits against:

```
72bc070  commit-author-email=julian@slushii.com  github-account=** UNLINKED **
933d11a  commit-author-email=julian@slushii.com  github-account=** UNLINKED **
```

If you add it under **Settings → Emails** (it can stay private), GitHub re-attributes retroactively and every commit you've made here shows up against your profile. Nothing needed on our side, and no history rewrite — which I'd rather not do to commits that are already tagged and correctly authored.

You're credited by name in [CONTRIBUTORS.md](https://github.com/sp00nznet/ps3recomp/blob/master/CONTRIBUTORS.md) and in the [v0.9.0 release notes](https://github.com/sp00nznet/ps3recomp/releases/tag/v0.9.0) either way, but the sidebar is the bit people actually look at.

PR #113, opened 2026-08-31
rsx: port the live NV4097 -> D3D12 draw engine, with HDR fixes

sp00nznet:
Brings caner's ([@canersaka](https://github.com/canersaka)) live draw engine across from the [Yakuza: Dead Souls EX](https://github.com/canersaka/Yakuza-Dead-Souls-EX) port so a second title can rasterise through it. `rsx_live_draw.c`, `rsx_dispatch.*`, `rsx_vertex_compact.*`, `rsx_restart_cuts.h` and the vp/fp decompiler updates are his work.

Opt-in behind `RSX_LIVE_DRAW` / `YZ_RSX_DRAW`, so the existing backend path is unchanged when it is off. `rsx_live_draw_config.c` gives the Yakuza-owned runtime config neutral zero definitions so the engine builds for a title that does not have them.

## Fixes found porting it to Twisted Metal (BCUS98106)

- **FIFO subchannel is a binding slot, not an engine selector.** This title binds NV4097 on subchannel 1, so `SET_SHADER_PROGRAM` was dropped and every draw ran one constant-black fragment program. Subchannels 0 and 1 both route to 3D now; `GCM_SUBCH1_2D=1` restores the old split.
- **HDR colour surfaces.** `SET_SURFACE_FORMAT` 0xB (`F_W16Z16Y16X16`) targets were created as `R8G8B8A8_UNORM`, clamping every value above 1.0 on write. Surface format now comes from the guest, the RTV format is part of the PSO key, and the debug dump tonemaps FP16 readback.
- **Fragment output NaN guard**, plus sanitising non-finite inline FP constants at both sites the decompiler reads them. A NaN fragment reads back as saturated white on FP16 and UNORM targets alike.
- **`W16Z16Y16X16_FLOAT` (0x9A) guest texture decode**, so a title's own FP16 render targets decode when sampled back as textures.

## Measured

Attract-mode run, before → after: texture binds falling back to the 1x1 white placeholder **30520 → 383** against 75133 real binds, and the FP16 render target no longer reads back as NaN.

## Known gaps

Surfaces still read back as a uniform clear colour at present time even though 25575 draw groups execute — likely a scissor/clip or dump-timing question, not the clamp/NaN causes fixed here. Three guest texture formats still fail to decode (`0xF2` 1472x1472, `0xBF` 1024x1, one malformed `0x00`).

https://claude.ai/code/session_01Q9SGujoEMo1hFpa6bMucQG


sp00nznet:
Pushed 12 more commits onto this branch: the SPURS event-flag work that takes YDKJ's CRI tasks from a hard deadlock to running. **All behavioural changes are opt-in; the default build is verified unchanged.**

**The chain**
- A PPU→SPU `EventFlagSet` could not reach a task that parked *without* registering a wait slot — our Set only signalled tasks in the flag's `used` bitmask, and these never register there. Same failure class @canersaka documented for SPURS queues in Yakuza Dead Souls (`f90e193`): "the HLE previously woke only its host condvar, which lifted SPU consumer tasks never wait on". Their fix reads a waiter record from the queue object; YDKJ publishes none, so this uses a runtime-side registry of parked tasks.
- The task reads its bits from **its own object at `+0x30`**, not the flag EA the PPU passed. The object sits at `flag - 0x80` and carries pointers to both its flags at `+0x14`/`+0x18`.
- Bits must be written at **both** Set time and consume time — neither ordering alone is safe. Healthy pattern went 1/4 → 8/10 runs: zero EXIT syscalls, all three tasks cycling `wait → work → wait`.
- `_cellSpursTaskAttributeInitialize` is a **six**-argument function, confirmed by two independently lifted titles (YDKJ + Jackbox Party Pack both set `r3..r8` and never write `r9`/`r10`). The old prototype read caller leftovers — `argument` came from `r10`, which in YDKJ is the ASCII `'.'` its hex-string parser stopped on.

**The headline, stated carefully**

`SPURS_EF_SPU_REPLY=2` forces the SPU→PPU flag the guest should set itself. With it YDKJ reaches **32 `DRAW_ARRAYS`**, a live present loop (3449 flips, alternating `bufferId`), compiled guest vertex/fragment programs, and `cellPad`/`cellHttp`/`cellSsl`.

**It is not a fix and there is no picture** — it fabricates the handshake, and every dumped frame is still pure black (`uniq=1, 000000`) even in runs logging 32 draws. What it proves is that the PPU's wait on `0x006B4600` is the single gate holding back the whole title.

**Next**: make the task set its own flag. `spu_func_000254D0` takes an early error exit when its argument's words 0 and 1 are both zero, and `gpr[89]` is built as `rotqmbyi(gpr[9], -4)` which always zeroes word 0 — so the guard hinges on the DMA'd context's word 0.

**Caveat for anyone testing**: this race is measurement-sensitive. Adding an `fprintf` probe changed which way it went, and enabling `YDKJ_DMATRACE` alone was enough to make the work path stop appearing. Never judge from a single run.

PR #133, opened 2026-09-05
cellSail: report async call completion to the title's event handler

sp00nznet:
`cellSailPlayerInitialize2` hands us a guest callback. We stored it in `callback_ea` and never called it — before this change the only two references to that field in the whole file were its declaration and that one assignment.

A title that drives SAIL asynchronously waits on that callback and never gets it. Tokyo Jungle's boot thread polls a flag word its own handler would set, every millisecond, for the life of the process — the flags stay `0x00000000` across thousands of polls.

### What the event actually is

`CellSailEvent` is `{be32 major; be32 minor}` packed into one 64-bit register, major in the high half. Async call completion is `major=2` (CALL_COMPLETED) with `minor` = the `CellSailPlayerCall` id that finished.

This wasn't guessed from an enum — it was read off the title's own handler (`func_0018195C`), which dispatches on the high half of r4:

| major | minor | handler does |
|---|---|---|
| 2 | 1 (BOOT) | set bit `0x1` |
| 2 | 2 (OPEN_STREAM) | set bit `0x10` |
| 2 | 10 (START) | clear `0x10000`, set `0x100` |
| 2 | 11 (STOP) | clear `0x100` |

Bit `0x1` is what its boot wait blocks on.

### Change

Adds `sail_notify(slot, major, minor, arg0)` and calls it at the end of `Boot`, `OpenStream`, `CloseStream`, `Start` and `Stop`. Every player call here runs synchronously, so the notify goes right after the work.

### Verified

Tokyo Jungle, before: main thread polls forever, flags never leave 0.
After: the poll loop exits, the title proceeds through its boot sequence and reaches `cellSaveDataAutoLoad2`.

Stacked on #130 (`fix/cellsail-player-handle`) since it touches the same file.

https://claude.ai/code/session_012g73rkhLETZBjZ7rG8Sw2h

### DIFF
--- a/libs/codec/cellSail.c
+++ b/libs/codec/cellSail.c
@@ -7,6 +7,7 @@
  * initial recompilation testing.
  */
 
+#include "ps3emu/guest_call.h"
 #include "cellSail.h"
 #include "../../runtime/ppu/ppu_memory.h"   /* vm_read32 / vm_write32 */
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


sp00nznet:
Landed — but not through this PR, so recording where it went.

This PR was stacked on #130 (`fix/cellsail-player-handle`). When #130 merged and its branch was deleted, GitHub auto-closed this one and will not reopen or retarget a PR whose base branch is gone.

The content is on `master` regardless: commit `9823d3a cellSail: report async call completion to the title's event handler` came in with the #130 merge, since the branches were stacked. Verified on master — `sail_notify` and the `callback_ea` plumbing are present in `libs/codec/cellSail.c`.

No action needed; noting it so the closed-not-merged state doesn't read as dropped work.

PR #135, opened 2026-09-06
cellSpurs: carry the SPU's answer to a job query in the completion event

sp00nznet:
A PPU-side query submits a job and reads the reply out of the completion event's **data3** (the guest stores r7 at sp+0xB8 and reads the value back from sp+0xBC).

`jc_signal_done` computed the SPU's outbound mailbox into `d2`/`d3` and then dropped it, pushing a hardcoded `0` — exactly what its own comment warned about:

> a synthetic payload means the caller reads back zero for whatever it asked

### Evidence

Tokyo Jungle asks each of its DSP modules for an input and an output bus count this way. Correlating each job's command block with the mailbox it posts:

```
cmd=0x0105 -> mbox=1      cmd=0x0106 -> mbox=1     <- the counts, answered correctly
cmd=0x0102 -> mbox=0x14860   cmd=0x0104 -> mbox=0x83C0   <- LS addresses, not counts
```

Reading back 0 leaves every module registered with zero buses, and the bus setup then rejects the first bus it is asked for — the check errors when `count < requested`, and `0 < 1`. With the answer forwarded, `1 < 1` is false and it passes.

### Scope

Only `0x105`/`0x106` are forwarded. The neighbouring commands reply with LS addresses, and `func_00201D38` stores each reply into `module+0x1C/+0x20/+0x24`. `data2` stays 0 because the audio consumer uses it as a **job index** — it waits until `data2 + 1` reaches the chain's job count, so a mailbox value there ends the wait early.

Adds `g_spurs_job_cmd` (the command read from the job's own command block) so the completion path can distinguish a query from ordinary work.

### Why it is off by default

`SPURS_ANSWER_QUERIES=1` to enable. Answering the query is correct, but the title then advances into a load that reads a ~23 MB file into a global buffer pointer at `[[TOC-0xB60]]` which nothing in the lifted image ever sets. The fread destination advances through the copy, so it walks from address 0 over the title's own OPD table and TOC. Until that NULL buffer is understood, the default keeps the clean reproducible failure rather than a corrupted one.

Full trail in `tokyojungle/docs/WHERE_IT_STOPS.md`.

https://claude.ai/code/session_012g73rkhLETZBjZ7rG8Sw2h

### DIFF
--- a/libs/codec/cellSail.c
+++ b/libs/codec/cellSail.c
@@ -7,6 +7,7 @@
  * initial recompilation testing.
  */
 
+#include "ps3emu/guest_call.h"
 #include "cellSail.h"
 #include "../../runtime/ppu/ppu_memory.h"   /* vm_read32 / vm_write32 */
 #include <stdio.h>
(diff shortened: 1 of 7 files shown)


sp00nznet:
Updated: the `SPURS_ANSWER_QUERIES` gate is **removed** — this is now on by default.

The gate existed because answering the query advanced Tokyo Jungle into a SPURS bring-up that failed, leaving a buffer unallocated so a later load read into NULL and walked over the title's own TOC. That bring-up failed for a separate reason: `sys_ppu_thread_once` had no implementation and reported success without running the initialiser (#136).

With #136 in place the allocation happens and the corruption is gone — a write-watch on the TOC slot records **zero** hits across runs. Tokyo Jungle now brings its audio up with no env knobs at all: 4 of 6 runs load its bus config plus menu/ranking/result/stage sound banks, against 2 files for the whole run before.

Depends on #136.

PR #177, opened 2026-09-10
SPU: HLE the taskset syscall for a standalone task on a planted context

sagemono:
SPU: HLE the taskset syscall for a standalone task on a planted context

A task dispatched on its own host thread runs on the SpursTasksetContext
spu_workload.c plants, whose syscallAddr is 0xA70, and nothing is lifted at
0xA70 for it. The intercept only fired for the legacy CRI image or in policy
mode, so LittleBigPlanet's audio tasks ended in the synthesised-stop exit on
their first syscall and the loading thread waited forever on their event
flag. Treat "context says 0xA70 and no lifted code at 0xA70" as the HLE case.
The syscall trace also prints the wait object.

### DIFF
--- a/runtime/spu/spu_channels.c
+++ b/runtime/spu/spu_channels.c
@@ -1529,8 +1529,9 @@ void spu_spurs_taskset_syscall(spu_context* ctx)   /* non-static: also called by
     uint32_t raw = ctx->gpr[3]._u32[0];
     uint32_t num = raw & 0x0F;
     { static int _n = 0; if (_n++ < 24)
-        fprintf(stderr, "[spu] SPURS taskset syscall num=%u (raw=0x%X args=0x%08X) image=%d link/r0=0x%05X\n",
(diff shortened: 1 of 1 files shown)


sp00nznet:
Merged the rest of the batch (#174, #175, #176, #178) and picked this one up — the Linux/macOS red was real, not flake.

`test_spu_lifted_start`'s "LLE taskset syscall executes resident policy code" sets `image_id = 16` with the policy registered in `resident_ovl = 84`, so `spu_lookup(0xA70, ctx->image_id)` found nothing and a real LLE taskset took the HLE path. Pushed a commit to this branch that asks the same images the dispatch below it would (`image_id`, `resident_task`, `resident_ovl`), which leaves your standalone case intact and puts the resident one back. Also folded the new wait-object field back onto one line — it carried an embedded `\n` and split every syscall trace in two.

CI re-running; merging once it is green. Thanks for the batch.

